PR #3, opened 2021-02-20
Restore block post to preserve external links

pablogsal:


gvanrossum (commented):
Thanks!

I know I could change the links, but I'd rather do that at some later time (I'm overwhelmed this weekend).

PR #5, opened 2021-02-22
Generalize annotation support with actions

isidentical:
Currently for annotations we can only use either a normal name (like `expr`) or a name followed by star (`expr_ty*`), though at least for python, type annotations are a bit more broad. Some examples;

```
start[ast.mod]: statements
strings[List[str]]: STRING+
```
These are currently unsupported, so what this PR does is actually unifies the annotation syntax with the mini-action grammar. And opens up the possibility of the using any sort of expression as types (also includes a little bugfix that allows you to create empty dictionary on the action part)

### DIFF
--- a/src/pegen/metagrammar.gram
+++ b/src/pegen/metagrammar.gram
@@ -57,8 +57,7 @@ rule[Rule]:
     | rulename memoflag? ":" alts NEWLINE { Rule(rulename[0], rulename[1], alts, memo=opt) }
 
 rulename[RuleName]:
-    | NAME '[' type=NAME '*' ']' { (name.string, type.string+"*") }
-    | NAME '[' type=NAME ']' { (name.string, type.string) }
+    | NAME annotation { (name.string, annotation) }
     | NAME { (name.string, None) }
 
 # In the future this may return something more complicated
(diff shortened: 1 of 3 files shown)


src/pegen/metagrammar.gram
```diff
@@ -112,19 +110,22 @@ atom[Plain]:
...
 
 target_atoms[str]:
     | target_atom target_atoms { target_atom + " " + target_atoms }
     | target_atom { target_atom }
 
 target_atom[str]:
-    | "{" ~ target_atoms "}" { "{" + target_atoms + "}" }
+    | "{" ~ atoms=target_atoms? "}" { "{" + (atoms or "") + "}" }
+    | "[" ~ atoms=target_atoms? "]" { "[" + (atoms or "") + "]" }
+    | NAME "*" { name.string + "*" }
```
isidentical, line 125: This is not actually needed, though makes it look pretty (`something * ` => `something*`)

gvanrossum:
Can you leave a comment explaining why you want this?

isidentical:
Uups, I thought I added it. Should be present in the description @gvanrossum

gvanrossum:
OK, got it. LGTM.

pablogsal:
@isidentical Could you add some tests to cover this?

lysnikolaou (approved):
Yup. LGTM!

pablogsal:
Thanks for the PR @isidentical :tada:

PR #10, opened 2021-03-17
Remove mentions of C from the docs and document the use of @ in grammar files

MatthieuDartiailh:
I also took the liberty to add support for @class that is showcased in examples but was not actually supported.


MatthieuDartiailh:
A quick question, some examples make use of the syntax `rule[type] (memo):` but the python parser generator does not seem to make use of that hint ? Should it ? It looks like everything is memoized anyway so I am not sure what is the purpose of this.

docs/grammar.md
```diff
@@ -111,13 +110,7 @@ rule_name[return_type]:
...
     | second_alt1 second_alt2 { second_alt1 }
 ```
-If the action is omitted and C code is being generated, then there
-are two different possibilities:
-1.  If there's a single name in the alternative, this gets returned.
-2.  If not, a dummy name object gets returned.
-
-If Python code is being generated, then a list with all the parsed
-expressions gets returned.
+If the action is omitted, a list with all the parsed expressions gets returned.
```
lysnikolaou, line 113: This should probably be changed after #14, right?

lysnikolaou (commented):
Thanks a lot for the PR, @MatthieuDartiailh! I've left one comment.

docs/grammar.md
MatthieuDartiailh: Indeed

MatthieuDartiailh: Addressed.

docs/grammar.md
```diff
@@ -128,6 +124,51 @@ The name can then be used in the action, like this:
 rule_name[return_type]: '(' a=some_other_rule ')' { a }
 ```
 
+
+### Customizing the generated parser
+
+By default, the generated parser inherits from the Parser class defined in pgen
+Parser, and is named GeneratedParser. One can customize the generated module by
```
lysnikolaou, line 131:
```suggestion
By default, the generated parser inherits from the Parser class defined in pegen/parser.py,
and is named GeneratedParser. One can customize the generated module by
```

docs/grammar.md
```diff
@@ -128,6 +124,51 @@ The name can then be used in the action, like this:
...
+modifying the header and trailer of the module generated by pegen. To do so one
+can add to the grammar dedicated sections discussed below:
```
lysnikolaou, line 133:
```suggestion
can add dedicated sections to the grammar, which are discussed below:
```

docs/grammar.md
```diff
@@ -128,6 +124,51 @@ The name can then be used in the action, like this:
...
+
+@class NAME
+
+This allows to specify the name of the generated parser.
+
+@header
+
+  Specify the header of the module as a string (one can typically use triple
```
lysnikolaou, line 141: The indentation here is not consistent with the other `meta`s. Let's have it consistent everywhere.

docs/grammar.md
lysnikolaou: Thanks.

docs/grammar.md
MatthieuDartiailh: I went with 2 space which makes for a nicer reading of the source IMO but does not cause the markdown to interprete it as code.

lysnikolaou (approved):
Thanks again @MatthieuDartiailh!

lysnikolaou:
@pablogsal I reviewed this, but leaving it open for you to have another look, since you assigned it to yourself.

pablogsal (approved):

pablogsal:
Sorry for the delay @MatthieuDartiailh  and thanks a lot for the PR! 🎉

Issue #6, opened 2021-02-27
Rule inlining


isidentical:
On an example grammar like this, 
```
@subheader """\
import ast
import itertools

def parse_number(token):
    for converter in [int, float, complex]:
        try:
            return converter(token.string)
        except ValueError:
            continue
    raise ValueError(f"Can't convert: {{number!r}}")

def parse_string(tokens):
    return ''.join(token.string for token in itertools.from_iterable(tokens))
"""

start[ast.mod]: a=expression NEWLINE* ENDMARKER { ast.dump(ast.Expression(a)) }

expression[ast.expr] (memo):
    | primary

primary[ast.expr]:
    | atom

atom[ast.expr]:
    | 'True' { ast.Constant(True) }
    | 'False' { ast.Constant(False) }
    | 'None' { ast.Constant(None) }
    | '...' { ast.Constant(Ellipsis) }
    | t_name
    | t_number
    | &STRING strings

t_name[ast.Name]: NAME { ast.Name(name.string, ast.Load()) }
t_number[ast.Constant]: NUMBER { ast.Constant(parse_number(number)) }

strings[ast.Constant] (memo): strings=STRING+ { ast.Constant(parse_string(strings)) }
```

When we have different rule groups (like `primary`, `expression` etc) this stuff gets piled up. So instead of seeing `Expression(body=Constant(value=1.0))` as the output, we see `Expression(body=[[[Constant(value=1.0)]]])`. This can be solved by adding actions that returns itself, like;
```
expression[ast.expr] (memo):
    | primary { primary }

primary[ast.expr]:
    | atom { atom }

atom[ast.expr]:
    | 'True' { ast.Constant(True) }
    | 'False' { ast.Constant(False) }
    | 'None' { ast.Constant(None) }
    | '...' { ast.Constant(Ellipsis) }
    | t_name { t_name }
    | t_number { t_number }
    | &STRING strings { strings }
```

or having an operator which actually does this under the hood. An example is the inline operator in the [lark-parser](https://lark-parser.readthedocs.io/en/latest/tree_construction.html#shaping-the-tree). This is how it might look
```
start[ast.mod]: a=expression NEWLINE* ENDMARKER { ast.dump(ast.Expression(a)) }

?expression[ast.expr] (memo):
    | primary

?primary[ast.expr]:
    | atom

?atom[ast.expr]:
    | 'True' { ast.Constant(True) }
    | 'False' { ast.Constant(False) }
[...]

PR #11, opened 2021-03-18
data: make a full Python parser for python grammar

MatthieuDartiailh:


MatthieuDartiailh:
I am making progress. But I have a couple of questions. 

First I had to include the fix to make the parser return a single item rather than a list of a single item. It is the first commit, I did not update the tests though. I will try to make a separate PR for that.

Second I am unsure how to make pegen happy regarding the start rule. It was manually set to file before but that does not seem completely correct what would be the right way to do this ?

Finally I would be interested in any strategy to test the parser and next how to properly handle line information.

pablogsal:
> Finally I would be interested in any strategy to test the parser and next how to properly handle line information.

For testing the parser, the best way is to throw it a lot of Python files (we used to have the whole stdlib or the top 5000 Pypi packages to do this) and compare potentially the output (or check if it crashes). If you plan to generate exactly the same AST, then it makes sense to compare the output with the built-in parser.

> Second I am unsure how to make pegen happy regarding the start rule. It was manually set to file before but that does not seem completely correct what would be the right way to do this ?

Not sure I fully understand the problem. I need some time to investigate exactly what's going on but it would be helpful if you could elaborate a bit more and maybe provide some examples :)



> First I had to include the fix to make the parser return a single item rather than a list of a single item

Could you also elaborate on this.

Currently, I am quite overburdened with work on CPython so apologies if it takes me a bit more than usually to respond here :)

MatthieuDartiailh:
> 
> > Finally I would be interested in any strategy to test the parser and next how to properly handle line information.
> 
> For testing the parser, the best way is to throw it a lot of Python files (we used to have the whole stdlib or the top 5000 Pypi packages to do this) and compare potentially the output (or check if it crashes). If you plan to generate exactly the same AST, then it makes sense to compare the output with the built-in parser.

I will look into how to do that then.

> 
> > Second I am unsure how to make pegen happy regarding the start rule. It was manually set to file before but that does not seem completely correct what would be the right way to do this ?
> 
> Not sure I fully understand the problem. I need some time to investigate exactly what's going on but it would be helpful if you could elaborate a bit more and maybe provide some examples :)

Pegen requires to have a rule start (or to have trailers). The Python grammar has multiple entries (file, interactive, eval). Looking at the original C grammar I realize now that the start is dynamically set in that case. I am not sure what would be the best way to emulate that, but for testing I can simply keep `start: file`

> 
> > First I had to include the fix to make the parser return a single item rather than a list of a single item
> 
> Could you also elaborate on this.

I was referring to the point discussed in #6 and for which I opened #14.

> 
> Currently, I am quite overburdened with work on CPython so apologies if it takes me a bit more than usually to respond here :)

MatthieuDartiailh:
I have reached a somehow puzzling point where trying to parse `import sys` lead to this being matched as an expression but then just bubbling up that way and at some point being converted to None. I tried to step through with pdb but I am not sure I fully follow the logic. I feel like the parser should realize that matching NAME won't go anywhere and then explore different options but it does not seem to be what happen. Parsing this appear to work though:
```python
def a():
    pass

a: int
a = 1
a += 1
a *= 1
a /= 1
a //= 1
a %= 1

a + 1
a * 1
a / 1
a // 1
a % 1

a < 1
a > 1
a <= 1
a >= 1
a == 1
a != 1

a = 2
a << 1
a >> 1
a ^ 1
a ^= 1

a = []
a = {}
```
but the ast test just checking the str complain about the context of some Name node being explicitely set to Load, I will look into making the ast comparison more robust at some point. 

Any help would be very much appreciated.

lysnikolaou:
At some point in time, we added hard keywords to the C parser, which means that keywords do not get parsed as NAMEs, but are themselves tokens. After that we moved the `star_expressions` alternative in the `simple_stmt` rule all the way to the top, in order to speed up the parsing, since expressions are the most common constructs in Python source code.

That's probably what's creating the problem. We have not implemented hard keywords for the python generator, which means that `import` gets parsed as a name and therefore as an expression. That's why `star_expressions` succeeds, way before `import_stmt` is even reached.

Does that make sense or have I missed something?

If you want to have a better look at how keywords are parsed in the C parser, have a look at the [keyword list](https://github.com/python/cpython/blob/3bb19873abd572879cc9a8810b1db9db1f704070/Parser/parser.c#L10) and [how it gets generated](https://github.com/python/cpython/blob/3bb19873abd572879cc9a8810b1db9db1f704070/Tools/peg_generator/pegen/c_generator.py#L455).

MatthieuDartiailh:
Thanks a lot @lysnikolaou ! Yes that make sense. Trying to browse through the C implementation am I right to assume that the key point is https://github.com/python/cpython/blob/3bb19873abd572879cc9a8810b1db9db1f704070/Parser/pegen.c#L629 which is in charge on transforming the tokens returned by the tokenizer and turn bare NAME token into the appropriate keyword TOKEN type.

lysnikolaou:
Exactly, yes! More specifically the function call to `_get_keyword_or_name_type` [here](https://github.com/python/cpython/blob/3bb19873abd572879cc9a8810b1db9db1f704070/Parser/pegen.c#L688).

MatthieuDartiailh:
Thanks for the confirmation.

MatthieuDartiailh:
I have now a better idea of how those things are connected but I still have a couple of questions:
- regarding soft keyword (I am not sure Python has any, but I would need them for my project), looking at the [c_generator](https://github.com/python/cpython/blob/3bb19873abd572879cc9a8810b1db9db1f704070/Tools/peg_generator/pegen/c_generator.py#L175) it looks like a soft keyword is not expected to be surrounded by `''`. Am I reading that properly and does that mean that `'import' NAME` would treat import as a keyword and `"import "NAME` would mean import is a soft keyword, since it has to be a string ?
- I am trying to figure out what is the best way to handle the token stream transformation. It seems natural to want to do that operation in the tokenizer. Pegen is quite tightly depending on the Python tokenizer which makes altering the tokenizer directly cumbersome (which is however something I wished could be done to add extra tokens, but I digress). The current approach is to wrap the bare Python tokenizer with a version that implement the proper methods, and handling the keyword transformation there would work fine. However I like the idea of extracting the keywords from the grammar to avoid duplicating things.  As a consequence, since the user is expected to create its own tokenizer, would it make sense to collect keywords on the parser and allow to initialize the tokenizer with a list of keywords that could be taken from the parser ?
- related to the previous one, should keyword be declared upfront or would identifying as we parse the grammar be fine ?

pablogsal:
> * regarding soft keyword (I am not sure Python has any, but I would need them for my project), looking at the [c_generator](https://github.com/python/cpython/blob/3bb19873abd572879cc9a8810b1db9db1f704070/Tools/peg_generator/pegen/c_generator.py#L175) it looks like a soft keyword is not expected to be surrounded by `''`. 

With pattern matching Python has `match` and `case` as soft keywords.

> Am I reading that properly and does that mean that `'import' NAME` would treat import as a keyword and `"import "NAME` would mean import is a soft keyword, since it has to be a string ?

Yup. Check for instance https://github.com/python/cpython/blob/master/Grammar/python.gram#L212

> * I am trying to figure out what is the best way to handle the token stream transformation. It seems natural to want to do that operation in the tokenizer. Pegen is quite tightly depending on the Python tokenizer which makes altering the tokenizer directly cumbersome (which is however something I wished could be done to add extra tokens, but I digress). The current approach is to wrap the bare Python tokenizer with a version that implement the proper methods, and handling the keyword transformation there would work fine. However I like the idea of extracting the keywords from the grammar to avoid duplicating things.  As a consequence, since the user is expected to create its own tokenizer, would it make sense to collect keywords on the parser and allow to initialize the tokenizer with a list of keywords that could be taken from the parser ?

Yeah, `pegen` is tightly coupled with the Python tokenizer and could be quite difficult to untangle since is baked into its base design. I think the best approach is to extract the keywords from the grammar and make the generated Python parser check for them as we are doing with the C parser as @lysnikolaou  mentioned. Maybe I am not understanding your proposal correctly, but if your concern is how to enforce keywords, then I think the best approach is that one. Check for example:

https://github.com/python/cpython/blob/af50c84643ce21cfbdfdabbdfae6bd5e1368c542/Tools/peg_generator/pegen/c_generator.py#L114-L134

> * related to the previous one, should keyword be declared upfront or would identifying as we parse the grammar be fine ?

They should be extracted from the grammar.

MatthieuDartiailh:
> 
> > * regarding soft keyword (I am not sure Python has any, but I would need them for my project), looking at the [c_generator](https://github.com/python/cpython/blob/3bb19873abd572879cc9a8810b1db9db1f704070/Tools/peg_generator/pegen/c_generator.py#L175) it looks like a soft keyword is not expected to be surrounded by `''`.
> 
> With pattern matching Python has `match` and `case` as soft keywords.

I started from the `python.gram` from the repo and pattern matching was not there yet so I missed that thanks for the pointer.

> 
> > Am I reading that properly and does that mean that `'import' NAME` would treat import as a keyword and `"import "NAME` would mean import is a soft keyword, since it has to be a string ?
> 
> Yup. Check for instance https://github.com/python/cpython/blob/master/Grammar/python.gram#L212

Thanks

> 
> > * I am trying to figure out what is the best way to handle the token stream transformation. It seems natural to want to do that operation in the tokenizer. Pegen is quite tightly depending on the Python tokenizer which makes altering the tokenizer directly cumbersome (which is however something I wished could be done to add extra tokens, but I digress). The current approach is to wrap the bare Python tokenizer with a version that implement the proper methods, and handling the keyword transformation there would work fine. However I like the idea of extracting the keywords from the grammar to avoid duplicating things.  As a consequence, since the user is expected to create its own tokenizer, would it make sense to collect keywords on the parser and allow to initialize the tokenizer with a list of keywords that could be taken from the parser ?
> 
> Yeah, `pegen` is tightly coupled with the Python tokenizer and could be quite difficult to untangle since is baked into its base design. I think the best approach is to extract the keywords from the grammar and make the generated Python parser check for them as we are doing with the C parser as @lysnikolaou mentioned. Maybe I am not understanding your proposal correctly, but if your concern is how to enforce keywords, then I think the best approach is that one. Check for example:
> 
> https://github.com/python/cpython/blob/af50c84643ce21cfbdfdabbdfae6bd5e1368c542/Tools/peg_generator/pegen/c_generator.py#L114-L134

My concern is that the token type need to be converted from NAME to the a dedicated type as is done in [_PyPegen_fill_token](https://github.com/python/cpython/blob/3bb19873abd572879cc9a8810b1db9db1f704070/Parser/pegen.c#L629) (along with other operations) and that feels like something that should be done inside the tokenizer rather than the parser, hence my idea to provide a way to do it on the tokenizer wrapper since we are already adding the ability to peek to the tokenizer. Ideally I would like to document well enough what are the requirements on the tokenizer to allow people to use a different one.

> 
> > * related to the previous one, should keyword be declared upfront or would identifying as we parse the grammar be fine ?
> 
> They should be extracted from the grammar.

Thanks

lysnikolaou:
> My concern is that the token type need to be converted from NAME to the a dedicated type as is done in _PyPegen_fill_token (along with other operations) and that feels like something that should be done inside the tokenizer rather than the parser, hence my idea to provide a way to do it on the tokenizer wrapper since we are already adding the ability to peek to the tokenizer.

Not doing this in the tokenizer was an explicit design decision though. While your feeling makes sense (we had this discussion when doing this), the general rule is that the tokenizer should know nothing about reserved keywords. Its only job is to identify different "sentence parts", without knowing about what they mean in the context they're parsed.

For example, `a.import()` should be tokenized as:
```
NAME a
DOT .
NAME import
LPAR (
RPAR )
```

It is up to the parser to identify that `import` is a reserved keyword and fail here.

MatthieuDartiailh:
Thanks for the clarification @lysnikolaou. Happy to know that, my feeling was not completely off. I will see how to implement the logic that way in the python parser/generator then. Thanks for your advice.

MatthieuDartiailh:
Good new I can now parse import statements ! 

However I would appreciate a quick review of the commit adding support for keywords because it is not without issues. Basically I followed the idea of the c generator and assigned new integers as token type. Even though I follow the idea that it should be done in the parser, I ended up doing inside the tokenizer wrapper since it seems the most efficient way to do it (this way the transformation is done once and only once). The keywords and soft keywords are collected from the grammar but due to how they collected are only listed at the very end of the generated parser. 

One issue I have is that printing anything involving TokenInfo fails because the type is unknown. Since we are basically doing string matching, I could use the OP type and still get the proper matching but it does not seems ideal. I am open to suggestions.

When I get more time I will go back to more extensively testing the parser.

MatthieuDartiailh:
I found a more elegant way to handle keyword by using the NAME matching rule. This way there is no need to touch the tokenizer and no modification of the token making them non printable.

Since to achieve this I had to make a number of changes not directly related to adding a full blown python grammar to the project I would like to know if you want me to split that part of and make a different PR for the Python parser.

Regarding testing for the Python parser, should I focus on the updating the scripts in scripts, or should I write more systematic tests under tests (maybe tests\python) ?

lysnikolaou:
> I found a more elegant way to handle keyword by using the NAME matching rule. This way there is no need to touch the tokenizer and no modification of the token making them non printable.

Happy to hear that.

> Since to achieve this I had to make a number of changes not directly related to adding a full blown python grammar to the project I would like to know if you want me to split that part of and make a different PR for the Python parser.

Truth is that would help a bunch with the review. If that's not too much work for you, I'd be really happy to review a PR with only the changes for keyword support.

> Regarding testing for the Python parser, should I focus on the updating the scripts in scripts, or should I write more systematic tests under tests (maybe tests\python) ?

I guess the answer is both. For now though, let's just focus on writing as many tests as possible for the constructs the parser can parse. For example, let's write tests that the output AST is correct for parsing import statements and then incrementally build more tests as we support more and more things. When we've got a first working version for all of the grammar, we can see if it makes sense to run modify and run something under `/scripts`.

MatthieuDartiailh:
Ok then I will try to make a new PR with the change to support for keyword, plus other minor stuff (such as not creating a method called async !), each with its own dedicated commit. And I will keep working on adding tests once I am done. 

Currently the one very broken things is the handling of strings. I need to look into it. I may try a trick similar to https://github.com/nucleic/enaml/blob/main/enaml/core/parser/parser36.py#L98 since we are parsing Python in Python and I do not fancy spending to much time on parsing f strings (but I could be convinced otherwise).

MatthieuDartiailh:
See #16 for a smaller PR for keyword support

MatthieuDartiailh:
I made some more progress (you can check the files under test/data to get an idea of what we can currently parse successfully). Right now I have some issue with parsing `f(a for a in b)` while `1 + f(a for a in b)` parse without issues. I checked the current python grammar and noticed that some rules make use of `&&` which is not currently the case. I will try to diff pegen vs the version currently in CPython main and see what has not yet been reflected in pegen.

MatthieuDartiailh:
After adding support for Forced node in the grammar definition and making some changes found in the current python grammar, I can now parse the tests files in the test directory and parse the source code of pegen ! I will try to add more tests and run on some popular pypi packages to get a better idea of how robust things are.

Next I will try to add pattern matching and line numbers in the AST. Regarding this last point I am interested into any suggestion to make this process less verbose and in ways to test the output since the repr of the AST will not be enough anymore, I imagine I will have to have test walking the AST in that case.

MatthieuDartiailh:
Note that black dislike tests files with Python 3.9+ only syntax. Should we exclude those files or bump the Python version running black ?

pablogsal:
Sorry for not participating here these days but I am very swamped with the beta release of Python 3.10. I just wanted to say that this effort actually could be very useful in the future to black, as we could add actions that create a CFG instead of an AST, allowing black to support Python 3.10+ syntax.

Thanks for all the patience @MatthieuDartiailh! Your work is great! Unfortunately, both @lysnikolaou and me are quite busy so it will take us a bit more to review 😅

MatthieuDartiailh:
No problem, I had to work on other projects impacted by changes in the latest alpha for a while and so I made only slow progresses.

MatthieuDartiailh:
I made some extra progress. I corrected a bunch of issues on async and decorators (I included a version check for the pre 3.9 syntax). python.gram based parser can now parse the 100 pypi packages tested but the included script ! If you are aware of additional corner cases worth testing, let me know.

As a side note, for async, I had to revert to identify async and await as keyword since tokenize send back NAME token not ASYNC token.

So I am left with pattern matching and line numbers to support. For the former, where can I find testing files, I assume some exist within CPython repo but I could not find them. For line numbers, I am open to any suggestion to make it efficient and on how to best test for correctness.

isidentical:
> So I am left with pattern matching and line numbers to support. For the former, where can I find testing files, I assume some exist within CPython repo but I could not find them. 

The file that we use for `ast.unparse` is [Lib/test/test_patma.py](https://github.com/python/cpython/blob/main/Lib/test/test_patma.py) (2000+ lines of pattern matching statements).

MatthieuDartiailh:
Thanks for the pointer @isidentical I will try to have look.

MatthieuDartiailh:
I did not look at more serious testing yet, but I tried to add line locations in the AST using a tactic similar to what is used in the c_generator. I also mirrored the most recent changes from cpython regarding error reporting but I still have more work to do to grab the sources and also to allow to select the entry point rule when creating a new parser instance.

As usual I am happy for any comment.

MatthieuDartiailh:
Regarding testing I usually use pytest, is that fine with you, before I add too much ? @pablogsal @lysnikolaou 

I have some basic testing working with correct line numbers. It took some patching for strings but I have no better idea at the moment. 

Now that the soft keyword PR is in let me know what feature would it make sense to extract next from this giant PR ?

MatthieuDartiailh:
While trying to add support for the last changes to python.gram I hit an issue with how to handle SOFT_KEYWORD. I can add a rule to the base parser class but then the grammar validator complains.

Should I add this to the tokens manually ? Also on the Python side we do not need the keys of the token.tok_name so I am wondering if this argument should not simply be a list or a set.

Some of the tests I added are not passing but I will look into this in the next days and try to add some coverage measurement to be able to add more tests.

As a side node @pablogsal, what is the purpose of the rule `invalid_with_stmt` which has no production ? I would have expected a syntax error.

pablogsal:
> As a side node @pablogsal, what is the purpose of the rule `invalid_with_stmt` which has no production ? I would have expected a syntax error.

Ah, that is using oke of the sneaky changes we made for syntax errors: the `&&` operator. That fails the parser in place if that token is not found without backtracking. Notice the rule is:

```
| [ASYNC] 'with' '(' ','.(expressions ['as' star_target])+ ','? ')' &&':'
| [ASYNC] 'with' ','.(expression ['as' star_target])+ &&':'
```

so is basically a way to emit a `expected ":"` error

pablogsal:
> @pablogsal thanks ! I guess I was mostly led astray by the fact that you have explicit rules to raise errors for missing : but also used force_expect in other places and then this rule that is kind of in between. Actually when writing tests I was confused by the fact that some construct had no error for missing `:` but errors for missing indent.
> 
> I think that using && for forced expectation is fine but it would need to be used in a more systematic manner to avoid confusion.

That cannot be done, when is not used is because stopping the parsing forcefully will lead to problems. That was the reason we ported usage of `&&` to the explicit checks in some cases. Specially when the parser goes into a left recursive rule, that can advance too much without being an error.

MatthieuDartiailh:
@pablogsal 

Is there is typo in the following rules and should it not be t_atom in the t_primary rule ? I am asking cause trying to improve coverage test of the parser I could not figure how to reach the target rule.
``` 
targets[asdl_expr_seq*]: a[asdl_expr_seq*]=','.target+ [','] { a }
target[expr_ty] (memo):
    | a=t_primary '.' b=NAME !t_lookahead { _PyAST_Attribute(a, b->v.Name.id, Store, EXTRA) }
    | a=t_primary '[' b=slices ']' !t_lookahead { _PyAST_Subscript(a, b, Store, EXTRA) }
    | t_atom
t_primary[expr_ty]:
    | a=t_primary '.' b=NAME &t_lookahead { _PyAST_Attribute(a, b->v.Name.id, Load, EXTRA) }
    | a=t_primary '[' b=slices ']' &t_lookahead { _PyAST_Subscript(a, b, Load, EXTRA) }
    | a=t_primary b=genexp &t_lookahead {
        _PyAST_Call(a, CHECK(asdl_expr_seq*, (asdl_expr_seq*)_PyPegen_singleton_seq(p, b)), NULL, EXTRA) }
    | a=t_primary '(' b=[arguments] ')' &t_lookahead {
        _PyAST_Call(a,
                 (b) ? ((expr_ty) b)->v.Call.args : NULL,
                 (b) ? ((expr_ty) b)->v.Call.keywords : NULL,
                 EXTRA) }
    | a=atom &t_lookahead { a }  <------------------- here
t_lookahead: '(' | '[' | '.'
t_atom[expr_ty]:
    | a=NAME { _PyPegen_set_expr_context(p, a, Store) }
    | '(' a=target ')' { _PyPegen_set_expr_context(p, a, Store) }
    | '(' b=[targets] ')' { _PyAST_Tuple(b, Store, EXTRA) }
    | '[' b=[targets] ']' { _PyAST_List(b, Store, EXTRA) }
```

pablogsal:
> Is there is typo in the following rules and should it not be t_atom in the t_primary rule ? I am asking cause trying to improve coverage test of the parser I could not figure how to reach the target rule.

I don't think so, if I remove it from the source:

```diff
diff --git a/Parser/parser.c b/Parser/parser.c
index d49bba1549..f38cc14595 100644
--- a/Parser/parser.c
+++ b/Parser/parser.c
@@ -17597,7 +17597,6 @@ target_rule(Parser *p)
 //     | t_primary '[' slices ']' &t_lookahead
 //     | t_primary genexp &t_lookahead
 //     | t_primary '(' arguments? ')' &t_lookahead
-//     | atom &t_lookahead
 static expr_ty t_primary_raw(Parser *);
 static expr_ty
 t_primary_rule(Parser *p)
```

i get this parsing failure:

```
 File "/home/pablogsal/github/python/master/./setup.py", line 29
    sys.modules['subprocess'] = _bootsubprocess
    ^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: invalid syntax
```

lysnikolaou:
Nope, it's not a typo. It was done by me in we-like-parsers/pegen_experiments#141 and if I recall correctly (it's been a while since) this aims to correctly handle the `ctx` argument to the ast node.

The rule without actions would just include `atom` as you correctly identified. However, because of the left-recursiveness of the thing we had to use two distinct rules. One for cases, where there is a single `Attribute` (or `Subscript`) node, which has to have a `Store` context and one for case where there are more nodes (like `a.b.c`), where the outermost node needs to be a `Store`, while the inner ones are `Load`s.

lysnikolaou:
> Thanks for the clarification. However I was not suggesting to remove it but to have t_atom instead of atom since otherwise t_atom and target appear to be unreachable to me or am I missing something else ?

That seems to be correct, yes. We could probably just remove `targets`, `target` and `t_atom` without further changes.

MatthieuDartiailh:
Thanks @pablogsal !

pablogsal:
> > Thanks for the clarification. However I was not suggesting to remove it but to have t_atom instead of atom since otherwise t_atom and target appear to be unreachable to me or am I missing something else ?
> 
> That seems to be correct, yes. We could probably just remove `targets`, `target` and `t_atom` without further changes.

Oh, excellent point @lysnikolaou and @MatthieuDartiailh !

MatthieuDartiailh:
On a similar note I believe the second rule of kwarg_or_double_starred is also unreachable since the first rule of kwargs will always match keywords arguments using the kwargs_or_starred rule. Am I right about this one ?

``` 
kwargs[list]:
    | a=','.kwarg_or_starred+ ',' b=','.kwarg_or_double_starred+ { a + b }
    | ','.kwarg_or_starred+
    | ','.kwarg_or_double_starred+
starred_expression:
    | '*' a=expression { ast.Starred(value=a, ctx=Load, LOCATIONS) }
kwarg_or_starred:
    | invalid_kwarg { UNREACHABLE }
    | a=NAME '=' b=expression { ast.keyword(arg=a.string, value=b, LOCATIONS) }
    | a=starred_expression { a }

kwarg_or_double_starred:
    | invalid_kwarg { UNREACHABLE }
    | a=NAME '=' b=expression { ast.keyword(arg=a.string, value=b, LOCATIONS) }   # XXX Unreachable
    | '**' a=expression { ast.keyword(arg=None, value=a, LOCATIONS) }
```

lysnikolaou:
Didn't exactly understand while rule you think is unreachable. Are you talking about a specific alternative of a rule?

MatthieuDartiailh:
My bad for some reason I had convinced myself that `f(a=1, **d, e=2)` was not legal... Sorry for the noise.

pablogsal:
Just to double check I have checked the C coverage over our parser once we remove the unused rules in https://github.com/python/cpython/pull/26655:

lel

MatthieuDartiailh:
Nice ! Happy I could help with this. And now I know that I should be able to test at least all functions.

MatthieuDartiailh:
I made some nice progress on testing (current coverage (with branch coverage enabled) is 81% without testing pattern matching). But I still have a bunch of bugs, in particular with syntax error checks, so I will keep working on that.

While going I made some well confined changes on the parser generation on which I would like to have your opinion:
- [support for SOFT_KEYWORD in rules](https://github.com/we-like-parsers/pegen/pull/11/commits/321edb5b4a97bafdb04c9f9eecd142fd335d62d8)
- [custom unreachable production for syntax error rules](https://github.com/we-like-parsers/pegen/pull/11/commits/a6dd362930b715298b50ff432f3c43acd354b155)
- [do not generate logic for cuts if no cuts are present. Those really cluttered the Python parser for no good reason](https://github.com/we-like-parsers/pegen/pull/11/commits/b065985b0fae7f97c1d1ca9bbf60c696050900aa)
- [line number and col offset tracking](https://github.com/we-like-parsers/pegen/pull/11/commits/981db88bf880297cfcc9f0fe7abe09467328c34f) and fixes in [1](https://github.com/we-like-parsers/pegen/pull/11/commits/268a3e1df08169ddd085a2fe2045c9d614a245fa), [2](https://github.com/we-like-parsers/pegen/pull/11/commits/1b95bd779e9283d1573e703d041b29589b2bb9e8), [3](https://github.com/we-like-parsers/pegen/pull/11/commits/08196d6a1ee5a14cb0a59fa34626026d850b6cc5)

For any of those I would be happy to extract the relevant changes in a separate PR and add the proper documentation, just let me know.

Finally a side question, currently the parser generator enforce that all grammar must have a start rule but for Python we have basically 5, and we just force start to be file. Would there be an interest in adding a way to declare that a grammar has multiple entry points ?

pablogsal:
> For any of those I would be happy to extract the relevant changes in a separate PR and add the proper documentation, just let me know.

Separate PRs (if possible) is always better :)

> Finally a side question, currently the parser generator enforce that all grammar must have a start rule but for Python we have basically 5, and we just force start to be file. Would there be an interest in adding a way to declare that a grammar has multiple entry points ?

Yeah, I think that may be interesting. Even better it would be the possibility of specifying the entry rule when building the parser or when calling "parse". That allows to test individual sub-trees rather elegantly.

Thanks a lot for the great work @MatthieuDartiailh!

Btw, the black maintainers are thinking of moving their parser to something else and I think they may be able to leverage this work. You may consider writing to them or adding some sentences in this document:

https://docs.google.com/document/d/1QPqm01E7MIRi_l4jrgCGOFVThP_HINnkm4hmVZC_CSg/edit?fbclid=IwAR03VBxQLhXkrZiO7BDUQerkeQU7MOy51BHtJ9xqRMHonBF7_dF1Tef53fs#

MatthieuDartiailh:
Thanks I commented on the google doc.

I will open the PRs when I get time but first I would like to track down a nasty bug. I have spent the better part of my evening on it but it fails to make sense somehow I believe the behavior I am seeing is right and I do not understand how it actually work in 3.10b2. The issue is as follows:

I am trying to parse `f(i for i in range(10))` which perfectly valid python. However I get a "generator expression must be parenthesized issue." Here is the overall parsing logic:
````
start() ... (looking at 1.0: NAME:'f')
  file() ... (looking at 1.0: NAME:'f')
    statements() ... (looking at 1.0: NAME:'f')
      _loop1_11() ... (looking at 1.0: NAME:'f')
        statement() ... (looking at 1.0: NAME:'f')
          compound_stmt() ... (looking at 1.0: NAME:'f') -> fails
          simple_stmts() ... (looking at 1.0: NAME:'f')
            simple_stmt() ... (looking at 1.0: NAME:'f')
              assignment() ... (looking at 1.0: NAME:'f')
                name() ... (looking at 1.0: NAME:'f')
                ... name() -> TokenInfo(type=1 (NAME), string='f', start=(1, 0), end=(1, 1), line='f(i for i in range(10))\n')
                expect(':') ... (looking at 1.1: OP:'(')
                ... expect(':') -> None
                _tmp_20() ... (looking at 1.0: NAME:'f') (i.e. '(' single_target ')' | single_subscript_attribute_target)
                expect('(') ... (looking at 1.0: NAME:'f')
                  ... expect('(') -> None
                single_subscript_attribute_target() ... (looking at 1.0: NAME:'f')
````
And inside `single_subscript_attribute_target` we have `t_primary`:
```
t_primary:
    | a=t_primary '.' b=NAME &t_lookahead { ast.Attribute(value=a, attr=b.string, ctx=Load, LOCATIONS) }
    | a=t_primary '[' b=slices ']' &t_lookahead { ast.Subscript(value=a, slice=b, ctx=Load, LOCATIONS) }
    | a=t_primary b=genexp &t_lookahead { ast.Call(func=a, args=[b], keywords=[], LOCATIONS) }
    | a=t_primary '(' b=[arguments] ')' &t_lookahead {
        ast.Call(
            func=a,
            args=b[0] if b else [],
            keywords=b[1] if b else [],
            LOCATIONS,
        )
     }
    | a=atom &t_lookahead { a }
t_lookahead: '(' | '[' | '.'
```
The first 2 options fail right away, the third identify the call with a generator as argument but bails for a missing look ahead, so logically we get to `a=t_primary '(' b=[arguments] ')' &t_lookahead` and here logically we complain that a non-parenthesized generator is invalid ! 

All of this seems perfectly logical but it is wrong and 3.10b2 which has the same grammar pass. So I am beyond confused. Actually removing the invalid argument check allow the code to pass since it now properly backtrack to the right point.

Any help or advice on how to debug that kind of issue will be greatly appreciated.

**Comment by MatthieuDartiailh** [comment]
After pondering on the explanation I just gave, I went to look at the c parser and realize that since cannot throw the exception only bubble if there is no alternative found. Looking back at my code I realized that I could not raise since that prevented the parser the backtrack properly, I am all good now. Sorry for the noise.

**Comment by MatthieuDartiailh** [comment]
Testing on 3.10 and after some addition to the tests, I managed to get 94% coverage. I add to juggle a bit with when to raise exception and when delay their raising, I hope I got it right. I would be curious to know how you handled that in the C parser. 

Also among the last things I tried to test were type comment, but it appears that tokenize does not expose TYPE_COMMENT. Is there any way to go around this issue. 

I will start putting together the PRs for the extra features I added to the generator.

**Comment by MatthieuDartiailh** [comment]
So all the small PRs are open ! and they are likely to conflict with one another at some point when merging. 

Once they are all in I will clean this one.

**Comment by MatthieuDartiailh** [comment]
@pablogsal @lysnikolaou any chance you could have a look at the small PRs I opened ?

**Comment by pablogsal** [comment]
> @pablogsal @lysnikolaou any chance you could have a look at the small PRs I opened ?

I have merged a bunch of them and left some comments in the rest. Thanks for the ping!

**Comment by cfbolz** [comment]
I just wanted to say thanks for all this work, we totally plan to steal it for PyPy once you're done ;-)

**Comment by MatthieuDartiailh** [comment]
You are welcome @cfbolz.If this helps PyPy to support Python 3.10 I am happy I could contribute in a sense.

Last commit include the changes from https://github.com/python/cpython/commit/0acc258fe6f0ec200ca2f6f9294adbf52a244802

**Comment by MatthieuDartiailh** [comment]
I "rebased" this branch on latest main. Here are what I believe are unresolved issues I would like to have some guidance on how best to address:
- how to retrieve source and lines in case of errors ? I feel like having to provide the path to the file as input to the parser would make the most sense. 
- how to best start the parsing ? Having a `parse` method taking a rule name would be the most flexible, but I have no strong opinion on this. I simply feel that having to have a `start` rule is a bit too rigid since Python grammar has basically 5 possible starts.
- how to handle the absence of TYPE_COMMENT in the tokens returned by the lexer ?
- add tests for extra input rules and add extra comments to the existing tests for the next people that will have to look into it.

Except the first point the others could be addressed in subsequent PR, I think.

**Comment by MatthieuDartiailh** [comment]
We also need to figure out how to best handle testing since we need test files containing invalid syntax or syntax only valid on python 3.9 or 3.10.

**Comment by pablogsal** [comment]
> I "rebased" this branch on latest main. Here are what I believe are unresolved issues I would like to have some guidance on how best to address:
> - how to retrieve source and lines in case of errors ? I feel like having to provide the path to the file as input to the parser would make the most sense. 

The easiest way is to store the string somewhere and reference that when showing errors. Anything else is an optimization. When parsing a file you can mimic the current exec () API where you pass the filename as well, but I would start simple.

Also I am a bit confused, we already show errors to some extent with the current python parser, although those were never very developed.

> - how to best start the parsing ? Having a `parse` method taking a rule name would be the most flexible, but I have no strong opinion on this. I simply feel that having to have a `start` rule is a bit too rigid since Python grammar has basically 5 possible starts.

We have the possibility to start from different rules in the c parser so it makes sense to have it in the Python one. Parse + rule method looks ok to me 

> - how to handle the absence of TYPE_COMMENT in the tokens returned by the lexer ?

I would suggest to not parse TYPE_COMMENT and remove them from the grammar. These are mainly useful for mypy and type checkers and otherwise they are comments in the source.

**Comment by pablogsal** [comment]
> We also need to figure out how to best handle testing since we need test files containing invalid syntax or syntax only valid on python 3.9 or 3.10.

AST based testing is only possible with the version of the syntax this is targeting. Unless you want to hardcode the expected AST or output, if you wanna compare the built in parser you need to run in the same version

**Comment by MatthieuDartiailh** [comment]
Regarding type comments anyway at this time we cannot do anything so far since the lexer does not propagate those. It however means the parser will be lossy (I have no idea if this may matter to black for example). Anyway it would require a change to CPython lexer itself I feel. For the time being I would keep the rules just to stay closer to CPython grammar which makes for easier comparison.

Regarding errors, I would need to store every single line since whenever CPython uses RAISE_SYNTAX_ERROR_KNOWN_LOCATION we only have access to AST nodes and hence line number but we do not have access anymore to the token which stores the line. I will go with storing all lines which will work for all inputs to the parser, and we could optimize later on for files.

Currently, one can start from any rule by simply calling the relevant method. But the grammar parser enforce to have a start rule. Should we first allow a grammar not to have a start rule ? Having a parse method does not bring much except if we were to allow to specify extra arguments along with the rule. I feel like this could be better designed in a follow up PR and we could limit ourselves to remove the need for a start rule.

Regarding testing, I am not concerned by the actual tests run by pytest but about the linting of the test files which cannot succeed. I will just make sure they are ignored.

**Comment by pablogsal** [comment]
> Regarding errors, I would need to store every single line since whenever CPython uses RAISE_SYNTAX_ERROR_KNOWN_LOCATION we only have access to AST nodes and hence line number but we do not have access anymore to the token which stores the line. I will go with storing all lines which will work for all inputs to the parser, and we could optimize later on for files.

Humm, unfortunately I don't follow. What I was mentioning is to store the full source if is available and otherwise store the text in some buffer if is buffered. Then, when reporting errors is just a matter of finding the appropriate line in said text. Is this what you are mentioning? In the C parser is very similar as we don't store the full source except in interactive mode, where there is no other way to reference errors

**Comment by MatthieuDartiailh** [comment]
The changes I did are in https://github.com/we-like-parsers/pegen/pull/11/commits/15a6363ff6469cdbf3ddef22b36ce13339af8be1. Basically I keep a mapping between line number and source by simply extracting from the tokens of the tokenizer. This is not optimal in term of memory since I have no clear idea when to purge this mapping but it should work fine.

**Comment by pablogsal** [comment]
> The changes I did are in https://github.com/we-like-parsers/pegen/pull/11/commits/15a6363ff6469cdbf3ddef22b36ce13339af8be1. Basically I keep a mapping between line number and source by simply extracting from the tokens of the tokenizer. This is not optimal in term of memory since I have no clear idea when to purge this mapping but it should work fine.

Hummm I can see how this works but I don't understand why we cannot just proceed in the same fashion we do in the c parser. Get the line number from the AST node on specific failures or from the last parsed token on general ones and just index a blob of text by line

**Comment by MatthieuDartiailh** [comment]
Now I am the one lost. I agree we have the line number from the AST for specific failures but where should one look up the source ? The CPython parser has some convoluted logic to find this blob (and in some cases even give up). For files from what I could follow it falls back to opening the fil.

**Comment by MatthieuDartiailh** [comment]
Another side question. I have encountered weird issues where the parser will fails if the file ends on multiple lines of comments (for example comment all but the first line of the assignments.py test file). The parser should work but does not and I don't grasp the difference with CPython parser that handles it just fine. Any pointer on this one ?

**Comment by MatthieuDartiailh** [comment]
If at one point you want to set up a phone/video call to go faster through the remaining issues/points to be discussed that would work for me.

**Comment by pablogsal** [comment]
> Now I am the one lost. I agree we have the line number from the AST for specific failures but where should one look up the source ? 

This is what I mentioned in my last comment:

> What I was mentioning is to store the full source if is available and otherwise store the text in some buffer if is buffered

If you are given a full string containing the source, store that. If you are given a buffered input, store the chunks in some buffer and join at the end when needed. I'm taking of the high level interface of the parser in which you either provide a string to be parsed or you provide a file to be parsed, in which case you can read it entirely and then store that.

Otherwise we can do something similar as you are doing, but we need to all agree on how do we want it and what are the disadvantages. @lysnikolaou what do you think?

>Another side question. I have encountered weird issues where the parser will fails if the file ends on multiple lines of comments (for example comment all but the first line of the assignments.py test file). The parser should work but does not and I don't grasp the difference with CPython parser that handles it just fine. Any pointer on this one ?

I can try to take a look at that next week as this one is the release of Python 3.10 beta 4 and things are unsurprisingly a bit on fire 🔥. Maybe @lysnikolaou had some time available to take a quick look before me.

Do you have a reproducer for this error?

**Comment by MatthieuDartiailh** [comment]
Here is a small reproducer:
````python
if a:
    b = 1

# ff


# r

```

I am trying to get all tests to pass also on Python 3.8 (there are a couple of trivial ast differences with attributes set to None). I am left with a single offset error coming from the presence of a ;

MatthieuDartiailh:
Cis are finally passing ! and on all targeted python versions. I will probably work on some other project till 3.10.b4 is released. Then we can discuss the exact API we want for the parser and how to complete this work.

pablogsal:
> If at one point you want to set up a phone/video call to go faster through the remaining issues/points to be discussed that would work for me.

That would simplify many things, we can try to schedule something for next week

pablogsal:
Please, ping me next week if I have not come back to you by that time in this PR.

MatthieuDartiailh:
Will do

MatthieuDartiailh:
ping @pablogsal

pablogsal:
> ping @pablogsal

@MatthieuDartiailh When is ok to have a chat? Does it work for you sometime tomorrow in the 11AM-5PM range London time?

MatthieuDartiailh:
We can say 1 PM London time. What platform would you like to use ?

pablogsal:
> We can say 1 PM London time. What platform would you like to use ?

I regularly use google meet because it requires less stuff to be installed, but I can do whatever. I can send a google calendar invite if you have a gmail account

MatthieuDartiailh:
Sounds good. My gmail is m.dartiailh@gmail.com (it is public on my github profile).

pablogsal:
> Sounds good. My gmail is [m.dartiailh@gmail.com](mailto:m.dartiailh@gmail.com) (it is public on my github profile).

Sent you an invite

MatthieuDartiailh:
Got it

MatthieuDartiailh:
As we discussed out of band, I added high level function handling the creation of the tokenizer and modified the retrieval of the source to avoid storing all lines for files. I did not add a function for interactive input since I am not sure what the signature should be.

By the way did you get a chance to look into https://github.com/we-like-parsers/pegen/pull/11#issuecomment-876666511 or do you have any suggestion on how to address the issue ? One we fix this issue I think this PR will be good to go.

pablogsal:
> By the way did you get a chance to look into https://github.com/we-like-parsers/pegen/pull/11#issuecomment-876666511 or do you have any suggestion on how to address the issue ? 

No, unfortunately I didn't had time to investigate. First we need to understand why is failing. You can instrument the python parser to print the tokens that rejects and accepts in the same fashion as we do in the C parser when you use `-d`. This may be related to something involving the end marker....

cfbolz:
> Here is a small reproducer:
> 
> ```python
> if a:
>     b = 1
> 
> # ff
> 
> 
> # r
> ```
> 
> I am trying to get all tests to pass also on Python 3.8 (there are a couple of trivial ast differences with attributes set to None). I am left with a single offset error coming from the presence of a ;

I cannot conclusively comment, but I was digging into a related issue a while ago, and what I found was that the C tokenizer that the pegen-parser in C uses has a slight difference in behaviour compared to the stdlib tokenize module, particularly related to newlines/ends of files: the stdlib pure python ones emits some extra tokens in some situations that the C tokenizer does not. And then the parser gets confused. So what I would try to do is print the tokens only for this example for both tokenizers and compare.

MatthieuDartiailh:
Thanks @cfbolz for the pointer that sounds like a reasonable direction. I tried to look up into the -d option to access the output of the C tokenizer but the documentation is rather sparse. Does anybody know how to access the output of the C tokenizer ?

pablogsal:
> Does anybody know how to access the output of the C tokenizer ?

There is nothing easy that lists emitted tokens currently (I am working on a parser-state-dump) so the only way is to instrument it yourself adding a bunch of `printf`

MatthieuDartiailh:
Great ! I guess I had to reach a point where I would have to recompile cpython in my python Journey. I will keep you posted.

cfbolz:
> 
> There is nothing easy that lists emitted tokens currently (I am working on a parser-state-dump) so the only way is to instrument it yourself adding a bunch of `printf` here: https://github.com/python/cpython/blob/7915c96ffd7ddc5cb6d54015ee4c31255a416892/Parser/pegen.c#L712

honestly, that's what I did, too :shrug: (lost the printf patch in the meantime though).

MatthieuDartiailh:
Patching cpython the token generated for:
```python
if a:
    b = 1

# ff
```
is:
```
Token 510: b'if'
Token 524: b'True'
Token 11: b':'
Token 4: b'' # NEWLINE
Token 5: b''  # INDENT
Token 1: b'b'
Token 22: b'='
Token 2: b'1'
Token 4: b''  # NEWLINE
Token 6: b''  # DEDENT
Token 0: b''
```

while tokenize.py yields:
```
TokenInfo(type=1 (NAME), string='if', start=(1, 0), end=(1, 2), line='if a:\n')
TokenInfo(type=1 (NAME), string='a', start=(1, 3), end=(1, 4), line='if a:\n')
TokenInfo(type=54 (OP), string=':', start=(1, 4), end=(1, 5), line='if a:\n')
TokenInfo(type=4 (NEWLINE), string='\n', start=(1, 5), end=(1, 6), line='if a:\n')
TokenInfo(type=5 (INDENT), string='    ', start=(2, 0), end=(2, 4), line='    b = 1\n')
TokenInfo(type=1 (NAME), string='b', start=(2, 4), end=(2, 5), line='    b = 1\n')
TokenInfo(type=54 (OP), string='=', start=(2, 6), end=(2, 7), line='    b = 1\n')
TokenInfo(type=2 (NUMBER), string='1', start=(2, 8), end=(2, 9), line='    b = 1\n')
TokenInfo(type=4 (NEWLINE), string='\n', start=(2, 9), end=(2, 10), line='    b = 1\n')
TokenInfo(type=61 (NL), string='\n', start=(3, 0), end=(3, 1), line='\n')
TokenInfo(type=60 (COMMENT), string='# ff', start=(4, 0), end=(4, 4), line='# ff')
TokenInfo(type=61 (NL), string='', start=(4, 4), end=(4, 4), line='# ff')
TokenInfo(type=4 (NEWLINE), string='', start=(4, 4), end=(4, 5), line='')
TokenInfo(type=6 (DEDENT), string='', start=(5, 0), end=(5, 0), line='')
TokenInfo(type=0 (ENDMARKER), string='', start=(5, 0), end=(5, 0), line='')
```

So as suspected tokenize does emit more tokens than the C tokenizer. Since we cannot fix tokenize on 3.8 and 3.9, we can either handle the issue in the tokenizer wrapper or in the parser. 
Is there any valid use of 2 successive NEWLINE (assuming we discard NL and COMMENT)? Naively I would say no but I am not sure. If it is indeed never valid we could filter it in the tokenizer wrapper otherwise we need some other heuristic in the parser.

pablogsal:
> 

> Token 510: b'if'
> Token 524: b'True'
> Token 11: b':'
> Token 4: b'' # NEWLINE
> Token 5: b''  # INDENT
> Token 1: b'b'
> Token 22: b'='
> Token 2: b'1'
> Token 4: b''  # NEWLINE
> Token 6: b''  # DEDENT
> Token 0: b''

What, why is emitting `Token 524: b'True'` ?

pablogsal:
> Is there any valid use of 2 successive NEWLINE (assuming we discard NL and COMMENT)? Naively I would say no but I am not sure. If it is indeed never valid we could filter it in the tokenizer wrapper otherwise we need some other heuristic in the parser.

I mean, not sure what you mean about "valid" but two NEWLINEs is absolutely possible, we cannot filter those.

 IIRC `NL` is emitted when there is an empty line:

```
❯ python -m tokenize <<< ""
1,0-1,1:            NL             '\n'
2,0-2,0:            ENDMARKER      ''
```

MatthieuDartiailh:
I meant two NEWLINE tokens (which are supposed to have meaning contrary to NL tokens). I will try to poke more around to see if I can find more odd cases and a rule to handle them.

MatthieuDartiailh:
> 
> 
> > 
> 
> > Token 510: b'if'
> > Token 524: b'True'
> > Token 11: b':'
> > Token 4: b'' # NEWLINE
> > Token 5: b''  # INDENT
> > Token 1: b'b'
> > Token 22: b'='
> > Token 2: b'1'
> > Token 4: b''  # NEWLINE
> > Token 6: b''  # DEDENT
> > Token 0: b''
> 
> What, why is emitting `Token 524: b'True'` ?

I just replaced `if a:` by `if True:` to avoid a name error at execution.

MatthieuDartiailh:
Playing a bit more with that it appears that the error only show up when there is a comment at the end of file without a blank line after it. In this case the tokenizer generate COMMENT, NL, NEWLINE, ENDMARKER, while in the presence of a newline it generates COMMENT, NL, ENDMARKER. If there is any DEDENT pending there are inserted before ENDMARKER but after the NEWLINE

So either the tokenizer wrapper needs to swallow that spurious NEWLINE, or we need [NEWLINE] in both the file rule and the block rule since the spurious new line appears before the DEDENT tokens. Any preference @pablogsal 

Should we also open a bug report on the python bug tracker too for this ? It does not look like a desirable behavior to me but I may be missing something.

pablogsal:
> Should we also open a bug report on the python bug tracker too for this ? It does not look like a desirable behavior to me but I may be missing something.

I think this looks like a tokenizer bug. I would recommend opening an issue so other folks can corroborate this or mentioned if this change is going to break anything for them

MatthieuDartiailh:
I opened the issue some time ago (https://bugs.python.org/issue44667) but it did not receive any comment. I will try to make a PR this week. In the meantime do you have any opinion @pablogsal @lysnikolaou  on how to handle the issue in older version (in the tokenizer or as extra rules) ? I personally prefer the second option to keep the grammar as close as possible to the existing one. This is the last blocker and after this PR can be merged. I will hence remove the WIP marker.

pablogsal:
@MatthieuDartiailh I agree that keeping the grammar as close as possible is probably the best course of action. I will try to investigate the bpo issue you opened after I release the first release candiate for 3.10 next week.

Otherwise, I would prefer to land this as soon as possible. Do you want to tacke the tokenizer issue in this PR or in another one?

pablogsal:
Also, @lysnikolaou if you have some spare minutes, it would be great if you could give the PR also a quick look.

lysnikolaou:
> Also, @lysnikolaou if you have some spare minutes, it would be great if you could give the PR also a quick look.

Yup! I'll make some time to give it a more thorough look tomorrow evening or Sunday morning.

MatthieuDartiailh:
I just pushed a fix for the tokenizer. Let me know what you think. I may not find time to back port your last 4 commits on CPython @pablogsal today but I will do my best. Feel free to merge without those.

pablogsal:
> I just pushed a fix for the tokenizer. Let me know what you think. I may not find time to back port your last 4 commits on CPython @pablogsal today but I will do my best. Feel free to merge without those.

What 4 commits are you referring to?

MatthieuDartiailh:
I have tried to apply any fix you made to the grammar file to the grammar file of this PR. The 4 commits I refer to are: 
https://github.com/python/cpython/commit/6948964ecf94e858448dd28eea634317226d2913
https://github.com/python/cpython/commit/ecc3c8e4216958d85385bf2467441c975128f26c
https://github.com/python/cpython/commit/302cf3550514b174b51b8c24d6e16cb47dc61755
https://github.com/python/cpython/commit/208a7e957b812ad3b3733791845447677a704f3e

pablogsal:
> I have tried to apply any fix you made to the grammar file to the grammar file of this PR. The 4 commits I refer to are: 
> https://github.com/python/cpython/commit/6948964ecf94e858448dd28eea634317226d2913
> https://github.com/python/cpython/commit/ecc3c8e4216958d85385bf2467441c975128f26c
> https://github.com/python/cpython/commit/302cf3550514b174b51b8c24d6e16cb47dc61755
> https://github.com/python/cpython/commit/208a7e957b812ad3b3733791845447677a704f3e

There is only one PR that you are probably interested which is a reordering of the rules for readability. The rules are the same, is just that they are grouped and commented so is not like a gigantic blob of text. The other PRs are for Syntax errors for prints, so is basically a single new rule.

pablogsal (approved):

pablogsal:
I will give some time for other folks to review and I will land afterwards. Then I will work on making a release

MatthieuDartiailh:
> 
> 
> > I have tried to apply any fix you made to the grammar file to the grammar file of this PR. The 4 commits I refer to are:
> > [python/cpython@6948964](https://github.com/python/cpython/commit/6948964ecf94e858448dd28eea634317226d2913)
> > [python/cpython@ecc3c8e](https://github.com/python/cpython/commit/ecc3c8e4216958d85385bf2467441c975128f26c)
> > [python/cpython@302cf35](https://github.com/python/cpython/commit/302cf3550514b174b51b8c24d6e16cb47dc61755)
> > [python/cpython@208a7e9](https://github.com/python/cpython/commit/208a7e957b812ad3b3733791845447677a704f3e)
> 
> There is only one PR that you are probably interested which is a reordering of the rules for readability. The rules are the same, is just that they are grouped and commented so is not like a gigantic blob of text. The other PRs are for Syntax errors for prints, so is basically a single new rule.

@pablogsal Should I make a new PR to address those ? I have the changes ready but I need to add a couple of tests.

pablogsal:
> @pablogsal Should I make a new PR to address those ? I have the changes ready but I need to add a couple of tests.

As you prefer. I would recommend at least doing https://github.com/python/cpython/commit/302cf3550514b174b51b8c24d6e16cb47dc61755

MatthieuDartiailh:
I did the re-organization and I added the new invalid rule but no test for it. Will do if I find time before this PR is merged.

lysnikolaou (approved):
Such great work! Thank you so much for going through with this, @MatthieuDartiailh!

MatthieuDartiailh:
Thanks @lysnikolaou !

pablogsal:
Fantastic! Thanks a lot, @MatthieuDartiailh for the hard work and the patience 🎉 🎉 🎉 

I will do a release after I manage to get 3.10.0 rc 1 out today 😉 

By the way, I fixed and backported https://bugs.python.org/issue44667

cfbolz:
nice, thanks everyone for your work on this!

PR #14, opened 2021-03-23
Do not wrap rule output in a list for rule consisting of a single item

MatthieuDartiailh:


pablogsal (approved):

pablogsal:
Thanks @MatthieuDartiailh !

PR #15, opened 2021-03-23
Do not format subheader

MatthieuDartiailh:
This allows to use Python formatting based on curly braces without having to double all curly braces.

### DIFF
--- a/src/pegen/python_generator.py
+++ b/src/pegen/python_generator.py
@@ -145,7 +145,7 @@ class PythonParserGenerator(ParserGenerator, GrammarVisitor):
             self.print(header.rstrip("\n").format(filename=filename))
         subheader = self.grammar.metas.get("subheader", "")
         if subheader:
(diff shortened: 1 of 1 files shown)


lysnikolaou:
Keep in mind that the docs will need to be updated for this after #10 is merged.

MatthieuDartiailh:
Will do yes

MatthieuDartiailh:
Actually I did not mention in #10 that the subheader was formatted (since I was planning on this PR going through), would you want to explicitly say it is not ?

lysnikolaou:
Oh right, that was the header. Never mind then.

MatthieuDartiailh:
Anything keeping this from being merged ?

pablogsal:
@MatthieuDartiailh Can you rebase? We can merge it afterwards

MatthieuDartiailh:
Done. Hopefully I got it right.

pablogsal (approved):

PR #16, opened 2021-04-08
Keyword support

MatthieuDartiailh:
Each commit should be easy to review separately (except for the _mark, _reset) but that is just automatic renaming. Mypy complains locally but for something also visible on main so not sure why we do not see it in CI.

I can add tests if necessary.

### DIFF
--- a/src/pegen/python_generator.py
+++ b/src/pegen/python_generator.py
@@ -54,7 +54,8 @@ class PythonCallMakerVisitor(GrammarVisitor):
             name = name.lower()
             return name, f"self.{name}()"
         if name in ("NEWLINE", "DEDENT", "INDENT", "ENDMARKER", "ASYNC", "AWAIT"):
-            return name.lower(), f"self.expect({name!r})"
+            # Avoid using names that can be Python keywords
+            return "_" + name.lower(), f"self.expect({name!r})"
         return name, f"self.{name}()"
 
     def visit_StringLeaf(self, node: StringLeaf) -> Tuple[str, str]:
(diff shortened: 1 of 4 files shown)


MatthieuDartiailh:
ping @lysnikolaou, if you have time for a review.

lysnikolaou:
Sorry @MatthieuDartiailh for taking so long with the review. I'll try to make some time until tomorrow evening.

MatthieuDartiailh:
ping in case somebody has time to give this a look.

lysnikolaou:
On it!

src/pegen/python_generator.py
```diff
@@ -47,17 +49,26 @@ class PythonCallMakerVisitor(GrammarVisitor):
...
         if name in ("NEWLINE", "DEDENT", "INDENT", "ENDMARKER", "ASYNC", "AWAIT"):
-            return name.lower(), f"self.expect({name!r})"
+            # Avoid using names that can be Python keywords
+            return "_" + name.lower(), f"self.expect({name!r})"
         return name, f"self.{name}()"
 
     def visit_StringLeaf(self, node: StringLeaf) -> Tuple[str, str]:
+        val = ast.literal_eval(node.value)
+        if re.match(r"[a-zA-Z_]\w*\Z", val):  # This is a keyword
+            if node.value.endswith("'") and node.value not in self.keywords:
```
lysnikolaou, line 68: No need for existence check, since we're working with sets.
```suggestion
            if node.value.endswith("'"):
```

lysnikolaou (commented):
Looks good @MatthieuDartiailh! Only left one relatively minor comment. Good job! :+1:

MatthieuDartiailh:
Done !

lysnikolaou (approved):
Yup!

PR #21, opened 2021-06-13
Add a special SOFT_KEYWORD rule allowing to match any soft keyword

MatthieuDartiailh:
This is used once in the Python parser for error reporting.

### DIFF
--- a/tests/test_pegen.py
+++ b/tests/test_pegen.py
@@ -562,8 +562,13 @@ def test_soft_keyword() -> None:
     grammar = """
     start:
         | "number" n=NUMBER { eval(n.string) }
-        | !(SOFT_KEYWORD) l=NAME n=NUMBER { f"{l.string} = {n.string}"}
+        | "string" n=STRING { n.string }
+        | SOFT_KEYWORD l=NAME n=(NUMBER | NAME | STRING) { f"{l.string} = {n.string}"}
     """
     parser_class = make_parser(grammar)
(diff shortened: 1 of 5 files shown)


tests/test_pegen.py
```diff
@@ -556,3 +556,14 @@ def test_missing_start() -> None:
     """
     with pytest.raises(GrammarError):
         parser_class = make_parser(grammar)
+
+
+def test_soft_keyword() -> None:
```
pablogsal, line 561: I am missing how this test is testing the `!(SOFT_KEYWORD)` part. As `"number" n=NUMBER` anything using `"number" NUMBER` will match the first rule without even testing the second one. You need to revert the order at least so the  `!(SOFT_KEYWORD)` is actually covered when you pass `"number 1"`, no?

I am missing something?

MatthieuDartiailh: I think you are right. I will double check.

pablogsal: I think a better test would be to have a bunch of rules with soft keywords and then have one that accepts SOFT_KEYWORD and something else and try that one with all the other soft keywords

MatthieuDartiailh: I did something along those lines. Let me know what you think.

pablogsal: Beautiful

pablogsal (approved):

PR #22, opened 2021-06-13
Add support for Forced grammar node in the parser generator

MatthieuDartiailh:
If the parsing fails, we immediately raise an error built using `make_syntax_error`

### DIFF
--- a/tests/test_pegen.py
+++ b/tests/test_pegen.py
@@ -584,7 +584,7 @@ def test_soft_keyword() -> None:
     with pytest.raises(SyntaxError):
         parse_string("test 1", parser_class, verbose=True)
 
-        
+
 def test_forced() -> None:
     grammar = """
     start: NAME &&':' | NAME
--- a/tests/test_pegen.py
+++ b/tests/test_pegen.py
@@ -597,7 +597,7 @@ def test_forced() -> None:
     assert "expected ':'" in str(e.exconly())
 
 
-def test_forced() -> None:
+def test_forced_with_group() -> None:
     grammar = """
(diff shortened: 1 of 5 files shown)


MatthieuDartiailh:
I have no idea what forced would mean on a group. Is this supposed to be supported ?

pablogsal:
> I have no idea what forced would mean on a group. Is this supposed to be supported ?

No, just on tokens

MatthieuDartiailh:
Currently this fails to generate a parser on group because my implementation of forced_expect calls on expect which cannot handle group. Do I get you right that I should fix that and allow to have things like `&&( a | b)` to work ? And a and b rule could be anything in that case.

MatthieuDartiailh:
I went with my last assumption and I have a passing test. Let me know if this is sufficient.

tests/test_pegen.py
```diff
@@ -583,3 +583,28 @@ def test_soft_keyword() -> None:
...
+    """
+    parser_class = make_parser(grammar)
+    assert parse_string("number :", parser_class, verbose=True)
+    with pytest.raises(SyntaxError) as e:
+        parse_string("a", parser_class, verbose=True)
+
+    assert "expected ':'" in str(e.exconly())
+
+
+def test_forced() -> None:
```
pablogsal, line 600: Only one of these two tests is running now because they are named the same 😉

MatthieuDartiailh: Sorry about that ! It is fixed.

pablogsal (approved):

pablogsal:
Fantastic!

PR #23, opened 2021-06-13
Add tracking to know if we are in a recursive rule

MatthieuDartiailh:
This is used for error reporting in the Python parser. I amunsure where to document this feature.

### DIFF
--- a/src/pegen/parser.py
+++ b/src/pegen/parser.py
@@ -111,7 +111,11 @@ def memoize_left_rec(method: Callable[[P], Optional[T]]) -> Callable[[P], Option
 
             while True:
                 self._reset(mark)
-                result = method(self)
+                self.in_recursive_rule += 1
+                try:
+                    result = method(self)
+                finally:
+                    self.in_recursive_rule -= 1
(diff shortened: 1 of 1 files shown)


pablogsal (approved):

src/pegen/parser.py
```diff
@@ -111,7 +111,11 @@ def memoize_left_rec_wrapper(self: P) -> Optional[T]:
 
             while True:
                 self._reset(mark)
-                result = method(self)
+                self.in_recursive_rule += 1
+                try:
+                    result = method(self)
+                finally:
+                    self.in_recursive_rule -= 1
```
pablogsal, line 118: This would look a bit better as a context manager, but I think this is fine in any case

pablogsal: If you want, create another PR for transforming this into a context manager, but is ok if you don't want to :)

MatthieuDartiailh: Since this shows up in a single place and is not meant to be used elsewhere I think the try finally is fine, otherwise I would agree.

PR #24, opened 2021-06-13
Add locations tracking inspired by the CPython implementation

MatthieuDartiailh:
This adds support for LOCATIONS in actions, which leads the parser to track linenos and col offsets for rules making use of it. We provide a way to customize how LOCATIONS is replaced in the generated parser.

### DIFF
--- a/src/pegen/python_generator.py
+++ b/src/pegen/python_generator.py
@@ -295,7 +295,7 @@ class PythonParserGenerator(ParserGenerator, GrammarVisitor):
                             action = f"{self.local_variable_names[0]}"
                         else:
                             action = f"[{', '.join(self.local_variable_names)}]"
-                elif "LOCATIONS" in node.action:
+                elif "LOCATIONS" in action:
                     self.print("tok = self._tokenizer.get_last_non_whitespace_token()")
                     self.print("end_lineno, end_col_offset = tok.end")
                     action = action.replace("LOCATIONS", self.location_formatting)
--- a/src/pegen/python_generator.py
+++ b/src/pegen/python_generator.py
@@ -1,7 +1,7 @@
 import ast
 import re
 import token
-from typing import IO, Any, Dict, Optional, Set, Text, Tuple
+from typing import IO, Any, Dict, Optional, Sequence, Set, Text, Tuple
 
 from pegen import grammar
 from pegen.grammar import (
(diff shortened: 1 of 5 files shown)


src/pegen/python_generator.py
```diff
@@ -176,6 +182,17 @@ def generate(self, filename: str) -> None:
...
 
+    def alts_uses_locations(self, alts: Sequence[Alt]) -> bool:
+        for alt in alts:
+            if alt.action and "LOCATIONS" in alt.action:
+                return True
+            for n in alt.items:
+                if isinstance(n.item, Alt) and n.item.action and "LOCATIONS" in n.item.action:
+                    return True
+                elif isinstance(n.item, Group) and self.alts_uses_locations(n.item.rhs.alts):
+                    return True
```
pablogsal, line 193: Could we add explicit individual tests for these branches?

src/pegen/tokenizer.py
```diff
@@ -60,6 +60,17 @@ def diagnose(self) -> tokenize.TokenInfo:
...
+    def get_last_non_whitespace_token(self) -> tokenize.TokenInfo:
+        index = self._index - 1
+        while index >= 0:
+            tok = self._tokens[index]
+            if tok.type != tokenize.ENDMARKER and (
+                tok.type < tokenize.NEWLINE or tok.type > tokenize.DEDENT
+            ):
+                break
+            index -= 1
+        return tok
```
pablogsal, line 72: You can probably simplify this using `itertools.islice` and a reversed iterator as you are basically iterating backwards and checking a condition on the given element.

MatthieuDartiailh: I pushed a change using reversed but I am not sure why I would need islice. 

Also could you please wait on this one for me to add some tests for the tokenizer specifically ?

pablogsal: > why I would need islice.

In theory to filter anything on the right of `self._index - 1`

> Also could you please wait on this one for me to add some tests for the tokenizer specifically ?

Sure

MatthieuDartiailh: Just slicing the list of _tokens directly as I did in my last commit should work just fine, no ? I may not have time to work on tests tonight (I am in France) but I will do my best to address those during the week.

pablogsal:
@MatthieuDartiailh please ping me when I can review this again

MatthieuDartiailh:
Will do.

MatthieuDartiailh:
@pablogsal I added a test both for LOCATIONS inside alt action and within group alt. I also added some minimal tests of the tokenizer. This is ready for review.

MatthieuDartiailh:
ping @pablogsal

pablogsal:
@MatthieuDartiailh You need to rebase on top of the other, PR. After that I think is ready :rocket:

MatthieuDartiailh:
Done @pablogsal

pablogsal:
Awesome, one less to go :)

MatthieuDartiailh:
I believe that was the last one !!! I will do my best to clean up the python parser one in the next days and summarize the open questions I have about it.

PR #25, opened 2021-06-20
Generate cut handling code only when necessary

MatthieuDartiailh:
I forgot this one last week...

### DIFF
--- a/tests/test_pegen.py
+++ b/tests/test_pegen.py
@@ -532,6 +532,17 @@ def test_cut() -> None:
     ]
 
 
+def test_cut_early_exit() -> None:
+    grammar = """
+    start: '(' ~ expr ')' | '(' name ')'
+    expr: NUMBER
+    name: NAME
+    """
+    parser_class = make_parser(grammar)
+    with pytest.raises(SyntaxError):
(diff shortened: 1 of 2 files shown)


tests/test_pegen.py
```diff
@@ -532,6 +532,17 @@ def test_cut() -> None:
     ]
 
 
+def test_cut_early_exit() -> None:
```
pablogsal, line 535: How does this test check that we are exiting early? Or is just for coverage?

pablogsal (approved):

tests/test_pegen.py
MatthieuDartiailh: If the cut was not honored we would succeed to parse `(a)` as `( NAME )` and the syntax error would not be raised.

pablogsal: Hummm, but the cut is honored even before this PR no? IIUC this is just optimizing when we add the error handling code

MatthieuDartiailh: Yes but at one point I messed up part of the logic and that was not working anymore. I think having a sanity check is a good thing.

PR #26, opened 2021-06-29
Fix implicit handling of unreachable

MatthieuDartiailh:
Alt using an invalid rule with no action should be marked automatically unreachable since their action will never be run. However previously we were marking as unreachable the alternative of invalid rules which is usually not useful and can be quite surprising for invalid_ helpers.

I noticed this issue while working on cleaning up the Python parser.

### DIFF
--- a/tests/test_pegen.py
+++ b/tests/test_pegen.py
@@ -626,7 +626,7 @@ def test_unreachable_explicit() -> None:
 def test_unreachable_implicit1() -> None:
     source = """
     start: NAME | invalid_input
-    invalid_input: NUMBER
+    invalid_input: NUMBER { None }
     """
     grammar = parse_string(source, GrammarParser)
     out = io.StringIO()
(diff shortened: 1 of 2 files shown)


pablogsal (approved):
I am a bit sad that we need a fully-fledged visitor just for this but it may be helpful in the future and I cannot see a more straightforward way to do this, so I am approving this,

tests/test_pegen.py
```diff
@@ -637,13 +637,25 @@ def test_unreachable_implicit1() -> None:
...
     invalid_input: NUMBER { None }
     """
     grammar = parse_string(source, GrammarParser)
     out = io.StringIO()
     genr = PythonParserGenerator(grammar, out, unreachable_formatting="This is a test")
     genr.generate("<string>")
+    assert "This is a test" in out.getvalue()
+
+
+def test_unreachable_implicit3() -> None:
```
pablogsal, line 650: I am assuming thsi tests covers all the weird things you saw that made you make this pr, but if there are more, please make a new PR with more tests :)

MatthieuDartiailh: This covers the cases that are actually used in the python grammar. I did not try to build contrived examples to go through all the branches of the visitor through. I will also check that my wording of the documentation is accurate.

PR #34, opened 2021-08-04
Fix the CI to test the correct versions

pablogsal:


.github/workflows/test.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10.0-beta.3']
```
MatthieuDartiailh, line 13: Don't you want to test on 3.10 rc 1 ? Since all the match rules are only tested on 3.10, it seemed better to me to test on 3.10 also.

pablogsal: Yeah, I plan to submit a new PR once the RC is available to test

MatthieuDartiailh: I believe it is already: https://github.com/actions/python-versions/releases

PR #36, opened 2021-08-05
cis: test on 3.10.0-rc.1

MatthieuDartiailh:


.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9']
+        python-version: ['3.8', '3.9', '3.10.0-rc.1']
```
pablogsal, line 16: You need to modify also the tox file, otherwise Isn't going to test with 3.10 IIRC

MatthieuDartiailh: Sorry went to fast I will fix that in a second.

MatthieuDartiailh: A side question why not run coverage on 3.9 and 3.10 ? At least for the parser it would make sense as otherwise we do not know anything about match coverage.

pablogsal: That's a very good point indeed, we should add coverage for 3.9 and 3.10. If you want to add it here that would be great, otherwise I can add it later. In the CI we should also add the flag to fail under a certain percentage of coverage

MatthieuDartiailh: I will add coverage measurement. Do you also want to upload coverage results to Codecov or another service ?

MatthieuDartiailh: And should all versions run with and without coverage ? I must say I do not see the point but I am not familiar with using tox.

pablogsal: In general I normally have both targets because coverage is slower and when you are iterating directly on it is a bit dragging but for the CI it makes sense to only run the coverage version, I think

MatthieuDartiailh: Ok so I will only run cov then. Do you want to upload to Codecov or you are not interested ?

MatthieuDartiailh:
I will be moving tomorrow so feel free to take over if you want.

pablogsal:
> I will be moving tomorrow so feel free to take over if you want.

Good luck with the moving 📦

.github/workflows/test.yml
pablogsal: > Do you want to upload to Codecov or you are not interested ?

Nah, I prefer to set up a minumum percentage once we are confortable with it. Thanks for the proposal!

PR #41, opened 2021-09-28
python.gram: reflect changes in cpython

MatthieuDartiailh:
See https://github.com/python/cpython/commit/e5f13ce5b48b551c09fdd0faeafa6ecf860de51c

I will try to add validation of the line and offsets values in the tests but maybe not before next week.

### DIFF
--- a/data/python.gram
+++ b/data/python.gram
@@ -1625,7 +1625,9 @@ invalid_arguments[NoReturn]:
      }
     | a=expression b=for_if_clauses ',' [args | expression for_if_clauses] {
         self.raise_syntax_error_known_range(
-            "Generator expression must be parenthesized", a, b[-1].target
+            "Generator expression must be parenthesized",
+            a,
+            (b[-1].ifs[-1] if b[-1].ifs else b[-1].iter)
         )
      }
     | a=NAME b='=' expression for_if_clauses {
@@ -1637,14 +1639,14 @@ invalid_arguments[NoReturn]:
         self.raise_syntax_error_known_range(
             "Generator expression must be parenthesized",
             a[0][-1],
-            b[-1].target
+            (b[-1].ifs[-1] if b[-1].ifs else b[-1].iter),
         ) if len(a[0]) > 1 else None
      }
     | args ',' a=expression b=for_if_clauses {
         self.raise_syntax_error_known_range(
             "Generator expression must be parenthesized",
             a,
-            b[-1].target,
+            (b[-1].ifs[-1] if b[-1].ifs else b[-1].iter),
         )
      }
     | a=args ',' args {
--- a/data/python.gram
+++ b/data/python.gram
@@ -154,9 +154,13 @@ class Parser(Parser):
                 f"{error_msg} is only supported in Python {min_version} and above."
             )
 
-    def raise_indentation_error(self, msg) -> None:
+    def raise_indentation_error(self, msg: str) -> None:
         """Raise an indentation error."""
-        raise IndentationError(msg)
+        last_token = self._tokenizer.diagnose()
+        args = (self.filename, last_token.start[0], last_token.start[1] + 1, last_token.line)
+        if sys.version_info >= (3, 10):
+            args += (last_token.end[0], last_token.end[1] + 1)
+        raise IndentationError(msg, args)
 
     def get_expr_name(self, node) -> str:
         """Get a descriptive name for an expression."""
--- a/data/python.gram
+++ b/data/python.gram
@@ -227,17 +227,17 @@ class Parser(Parser):
         node.ctx = context
         return node
 
-    def ensure_real(self, number_str: str):
-        number = ast.literal_eval(number_str)
-        if type(number) is complex:
-            self.raise_syntax_error("real number required in complex literal")
-        return number
+    def ensure_real(self, number: ast.Constant):
+        value = ast.literal_eval(number.string)
+        if type(value) is complex:
+            self.raise_syntax_error_known_location("real number required in complex literal", number)
+        return value
 
-    def ensure_imaginary(self, number_str: str):
-        number = ast.literal_eval(number_str)
-        if type(number) is not complex:
-            self.raise_syntax_error("imaginary number required in complex literal")
-        return number
+    def ensure_imaginary(self, number:  ast.Constant):
+        value = ast.literal_eval(number.string)
+        if type(value) is not complex:
+            self.raise_syntax_error_known_location("imaginary number required in complex literal", number)
+        return value
 
     def generate_ast_for_string(self, tokens):
         """Generate AST nodes for strings."""
@@ -986,10 +986,10 @@ signed_real_number:
     | '-' real=real_number { ast.UnaryOp(op=ast.USub(), operand=real, LOCATIONS) }
 
 real_number[ast.Constant]:
-    | real=NUMBER { ast.Constant(value=self.ensure_real(real.string), LOCATIONS) }
+    | real=NUMBER { ast.Constant(value=self.ensure_real(real), LOCATIONS) }
 
 imaginary_number[ast.Constant]:
-    | imag=NUMBER { ast.Constant(value=self.ensure_imaginary(imag.string), LOCATIONS) }
+    | imag=NUMBER { ast.Constant(value=self.ensure_imaginary(imag), LOCATIONS) }
 
 capture_pattern:
     | target=pattern_capture_target {
--- a/data/python.gram
+++ b/data/python.gram
@@ -373,17 +373,28 @@ class Parser(Parser):
         message: str,
         start: Optional[Tuple[int, int]] = None,
         end: Optional[Tuple[int, int]] = None
-    ) -> None:
+    ) -> NoReturn:
         raise self._build_syntax_error(message, start, end)
 
     def make_syntax_error(self, message: str) -> None:
         return self._build_syntax_error(message)
 
-    def raise_syntax_error(self, message: str) -> None:
-        """Raise a syntax error."""
-        raise self._build_syntax_error(message)
+    def expect_forced(self, res: Any, expectation: str) -> Optional[tokenize.TokenInfo]:
+        if res is None:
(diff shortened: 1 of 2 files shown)


data/python.gram
```diff
@@ -1580,10 +1580,10 @@ invalid_arguments[NoReturn]:
             "invalid syntax. Maybe you meant '==' or ':=' instead of '='?", a, b
         )
      }
-    | a=args for_if_clauses {
-        self.store_syntax_error_starting_from(
+    | a=args b=for_if_clauses {
+        self.store_syntax_error_known_range(
             "Generator expression must be parenthesized",
-            a[1][-1] if a[1] else a[0][-1]
+            a[1][-1] if a[1] else a[0][-1], b
```
pablogsal, line 1586: Hummm, not sure this works, you need to get the node that is further away in the comprehension. Check the c code helper in the CPython repo

MatthieuDartiailh: Yeah I realized based on the failed checks, I need b[-1], I think I will test it before pushing this time.

data/python.gram
```diff
@@ -303,10 +303,10 @@ class Parser(Parser):
                 self._tokenizer.get_lines(list(range(start[0], end[0] + 1)))
             )
 
-        self._exception = SyntaxError(
-            message,
-            (self.filename, start[0], start[1], line)
-        )
+        args = (self.filename, start[0], start[1], line)
+        if sys.version_info >= (3, 10):
+            args += (stop[0], stop[1])
```
pablogsal, line 308: I assume this is for printing the new ranged errors in Python 3.10, no?

tests/python_parser/test_syntax_error_handling.py
```diff
@@ -22,8 +24,17 @@ def parse_invalid_syntax(python_parse_file, python_parse_str, tmp_path, source,
...
-    assert message in str(e.exconly())
+    print(str( e.exconly()))
+    assert message in str( e.exconly())
+
+    exc = e.value
+    if start:
+        assert exc.lineno == start[0]
+        assert exc.offset == start[1]
+
+    if stop and sys.version_info >= (3, 10):
```
pablogsal, line 35: Nice!

data/python.gram
MatthieuDartiailh: yes

MatthieuDartiailh: And because I realize when trying to add testing that it was missing.

tests/python_parser/test_syntax_error_handling.py
MatthieuDartiailh: Thanks. However adding actual test values will take me some time. Feel free to add some yourself if you want.

pablogsal:
@MatthieuDartiailh You need to rebase this branch on top of main now that #60 has landed.

MatthieuDartiailh:
Will do yes and I need to address the question of the tests too.

MatthieuDartiailh:
All tests pass locally on 3.10.0, I will try to fix the three broken tests when I get a chance but I would already appreciate a review.

MatthieuDartiailh:
The three failing are related to this rule
```
| !(NAME STRING | SOFT_KEYWORD) a=disjunction b=expression_without_invalid {
        _PyPegen_check_legacy_stmt(p, a) ? NULL : p->tokens[p->mark-1]->level == 0 ? NULL :
        RAISE_SYNTAX_ERROR_KNOWN_RANGE(a, b, "invalid syntax. Perhaps you forgot a comma?") }
```
but tokenize.py does not trac the level so emulating this will be tricky.

pablogsal:
> The three failing are related to this rule
> ```
> | !(NAME STRING | SOFT_KEYWORD) a=disjunction b=expression_without_invalid {
>         _PyPegen_check_legacy_stmt(p, a) ? NULL : p->tokens[p->mark-1]->level == 0 ? NULL :
>         RAISE_SYNTAX_ERROR_KNOWN_RANGE(a, b, "invalid syntax. Perhaps you forgot a comma?") }
> ```
> but tokenize.py does not trac the level so emulating this will be tricky.

Yeah, we can ignore it for the time being. We could try to reformulate it using existing information or remove the restriction for now and let it be more noisy.

MatthieuDartiailh:
Once this go in I will work on adding the latest improvement to syntax errors:
bpo45716
bpo45764
bpo45727
bpo45450
bpo46836

pablogsal:
Is this now ready for review?

MatthieuDartiailh:
Yes

MatthieuDartiailh:
ping @pablogsal @lysnikolaou 

Would it it be possible to get a review for this ?

pablogsal:
> ping @pablogsal @lysnikolaou 
> 
> Would it it be possible to get a review for this ?

Yeah, I will try to get to this this week. As we are close to 3.11b1 I'm getting a ton of extra work these weeks on CPython so I am a bit overwhelmed.

Apologies for the delay :(

data/python.gram
```diff
@@ -171,59 +189,86 @@ class Parser(Parser):
                 f"(line {node.lineno})."
             )
 
+    def get_invalid_target(self, target: Target, node: Optional[ast.AST]) -> Optional[ast.AST]:
```
pablogsal, line 192: Note to self: this mirrors `_PyPegen_get_invalid_target`

data/python.gram
```diff
@@ -171,59 +189,86 @@ class Parser(Parser):
...
+            err = e
 
         # Avoid getting a triple nesting in the error report that does not
         # bring anything relevant to the traceback.
-        if err_msg:
-            self.raise_syntax_error_known_location(err_msg, t)
-            raise self._exception
-
-        return m.body[0].value
+        if err is not None:
```
pablogsal, line 263: This keeps the exception alive more than it should, and this will delay the destruction of the frame due to it forming a cycle with the frame. Let's make sure we delete de reference explicitly when we don't need it anymore by adding `del err` or by changing this code to capture only the message (as it was previously).

data/python.gram
```diff
@@ -171,59 +189,86 @@ class Parser(Parser):
...
+            args = (err.filename, err.lineno + line_offset - 2, err.offset, err.text)
+            if sys.version_info >= (3, 10):
+                args += (err.end_lineno + line_offset - 2, err.end_offset)
+            raise SyntaxError(err.msg, args)
+
+        node = m.body[0].value
+        ast.increment_lineno(node, line_offset - 2)
```
pablogsal, line 270: Why do we need to do this?

data/python.gram
```diff
@@ -317,27 +362,39 @@ class Parser(Parser):
                 self._tokenizer.get_lines(list(range(start[0], end[0] + 1)))
             )
 
-        return  SyntaxError(
-            message,
-            (self.filename, start[0], start[1], line)
-        )
+        args = (self.filename, start[0], start[1] + 1, line)
```
pablogsal, line 365: Can we add a comment on why now we have an offset of 1?

data/python.gram
```diff
@@ -317,27 +362,39 @@ class Parser(Parser):
...
         end: Optional[Tuple[int, int]] = None
-    ) -> None:
+    ) -> NoReturn:
         raise self._build_syntax_error(message, start, end)
 
     def make_syntax_error(self, message: str) -> None:
         return self._build_syntax_error(message)
 
-    def raise_syntax_error(self, message: str) -> None:
+    def expect_forced(self, res: Any, expectation: str) -> Optional[tokenize.TokenInfo]:
```
pablogsal, line 382: 👍

pablogsal, line 382: Where are we using this method? IIRC forced tokens already work (https://github.com/we-like-parsers/pegen/blob/e28fe4fb1972c55af5ddb6a7bdd9cba4ea072b81/tests/test_pegen.py#L589)

data/python.gram
MatthieuDartiailh: This would mean losing the exact matching to CPython (which I worked hard to get). I would prefer adding an explicit delete even though I do not see much difference since we can only delete it right before raising.

data/python.gram
MatthieuDartiailh: I simplified the code to avoid generating many blank line so this is my way to get back the right lineno.

data/python.gram
MatthieuDartiailh: Will do. This is related to @edemaine comment that tokenize index at 0 and SyntaxError at 1.

data/python.gram
MatthieuDartiailh: This is needed to get the right error location. CPython report equal start and end for forced token which is not what we do in the default implementation. The default parser only has make_syntax_error which queries the last token and use start and end which is reasonable in general but not in this special case.

data/python.gram
pablogsal: > delete it right before raising.

That makes all the difference because it avoids the cycle that will be propagated as the frame in the exception bubbles up. This is anyway a small detail, so don't worry too much about it if you want to leave as is :)

data/python.gram
pablogsal: Do we have tests covering this difference? Maybe we should add some to `test_pegen.py` to be explicit about it.

MatthieuDartiailh: We have 4 tests in test_syntax_error_handling.py failing if I comment this out.

data/python.gram
MatthieuDartiailh: Added a comment

data/python.gram
MatthieuDartiailh: Added a comment

MatthieuDartiailh:
ping @pablogsal

pablogsal (approved):
LGTM

Thanks a lot, @MatthieuDartiailh for the patience and for the fantastic work. I know how much this work takes and I wanted to highlight how awesome is that you dedicated a lot of effort to get parity with the latest changes.

I apologized for the time this has been lying around, but the release of 3.11 is proving to be challenging 😅

MatthieuDartiailh:
Thanks @pablogsal !

#64 should be quite easy to review and add next. My other 2 PRs require some more discussions.

PR #43, opened 2021-09-28
Add a first version of the docs

pablogsal:


pablogsal:
CC: @edemaine

docs/grammar.rst
```diff
@@ -0,0 +1,256 @@
...
+
+::
+
+     rule1: rule2 | 'a'
+     rule2: rule3 | 'b'
+     rule3: rule1 | 'c'
+
+and “hidden left-recursion” like::
+
+::
```
edemaine, line 170: This doesn't render correctly (two `::`s).

docs/grammar.rst
```diff
@@ -0,0 +1,256 @@
...
+     rule: 'optional'? rule '@' some_other_rule
+
+Variables in the Grammar
+------------------------
+
+A sub-expression can be named by preceding it with an identifier and an
+``=`` sign. The name can then be used in the action (see below), like
+this: ::
+
+::
```
edemaine, line 181: Ditto here.

docs/grammar.rst
```diff
@@ -0,0 +1,256 @@
...
+
+       rule_name: ','.e+
+
+.. _e-3:
+
+``&e``
+~~~~~~
+
+Succeed if e can be parsed, without consuming any input.
+
```
edemaine, line 125: One thing missing here, and in much documentation of pegen that I've seen (PEP etc.), is what `&&` means. It's used a bunch in cpython and data's `python.gram`.

Hmm, but somehow it's documented in `grammar.md`. Should that redundant file be removed?

(I tried to look for other such undocumented symbols, but I don't see any.)

docs/grammar.rst
```diff
@@ -0,0 +1,256 @@
...
+code inside curly-braces, which specifies the return value of the
+alternative
+
+::
+
+       rule_name[return_type]:
+           | first_alt1 first_alt2 { first_alt1 }
+           | second_alt1 second_alt2 { second_alt1 }
+
+If the action is ommited, a default action is generated:
```
edemaine, line 215:
```suggestion
If the action is omitted, a default action is generated:
```

docs/grammar.rst
```diff
@@ -0,0 +1,256 @@
...
+
+-  If there’s a single name in the rule in the rule, it gets returned.
+
+-  If there is more than one name in the rule, a collection with all
+   parsed expressions gets returned.
+
+This default behaviour is primarily made for very simple situations and
+for debugging pourposes.
```
edemaine, line 223:
```suggestion
for debugging purposes.
```

docs/grammar.rst
```diff
@@ -0,0 +1,256 @@
...
+In the specific case of the PEG grammar for Python, having actions
+allows directly describing how the AST is composed in the grammar
+itself, making it more clear and maintainable. This AST generation
+process is supported by the use of some helper functions that factor out
+common AST object manipulations and some other required operations that
+are not directly related to the grammar.
+
+To indicate these actions each alternative can be followed by the action
+code inside curly-braces, which specifies the return value of the
+alternative
```
edemaine, line 207: One thing that confused me at first, so could use documentation here, is that the thing in braces must be a Python expression; it can't be multiple lines of Python code.  This is sort of implied by it being `return`ed, but I think it would help to be explicit. Another possible interpretation (and possibly a useful feature in the future) would be to allow multiple statements which include a `return`.

Also missing a period.

edemaine (commented):
Looking nice! Thanks for deploying to gh-pages so I could easily look at a rendered version.

docs/grammar.rst
pablogsal: > , is what `&&` means. It

Ah, excellent point. This was the only grammar element that was added **after** the PEP. I will make sure to add it there.

> Hmm, but somehow it's documented in `grammar.md`. Should that redundant file be removed?

Indeed, I will merge these two files.

pablogsal: I pushed a change merging both files

edemaine: Great! Some important other additions here, like documenting `LOCATIONS`, `@class`, and `@subheader`.  I added a few specific comments to [the commit](https://github.com/we-like-parsers/pegen/commit/0328de304e19fd931726d213fbc88b5a2e1dab68).

pablogsal:
Thanks, @edemaine for the great review. I have pushed a new commit addressing the feedback.

docs/grammar.rst
edemaine: I think this comment is the only one remaining.  Here's a possible wording:

```
To indicate these actions, each alternative can be followed by an action
inside curly-braces, which specifies a Python expression to be evaluated
and returned for the alternative:
```

docs/grammar.rst
```diff
@@ -0,0 +1,395 @@
...
+In the specific case of the PEG grammar for Python, having actions
+allows directly describing how the AST is composed in the grammar
+itself, making it more clear and maintainable. This AST generation
+process is supported by the use of some helper functions that factor out
+common AST object manipulations and some other required operations that
+are not directly related to the grammar.
+
+To indicate these actions each alternative can be followed by the action
+code inside curly-braces, which specifies the return value of the
+alternative:
```
pablogsal, line 241:
```suggestion
To indicate these actions, each alternative can be followed by an action
inside curly-braces, which specifies a Python expression to be evaluated
and returned for the alternative:
```

PR #44, opened 2021-09-28
Sort KEYWORDS for consistent output ordering

### LINKED ISSUE #40: Consistent keyword ordering
Every time I run pegen on the same grammar, I (often) get a different order for `KEYWORDS` and `SOFT_KEYWORDS`.  This is annoying for version control, when `parse.py` is committed to a repo.

I imagine the cause is that the keywords are stored in a `set` which has nondeterministic order, and `sorted` will make this behavior deterministic.  Specifically, `tuple(...)` should be replaced with `tuple(sorted(...))` in these lines:

https://github.com/we-like-parsers/pegen/blob/31f489a16ba961be2b7142789f54e4a02561acbc/src/pegen/python_generator.py#L240-L241

Is it appropriate to make PR here, or would it be better against cpython?

edemaine:
Fixes #40 by adding `sorted(...)` as suggested previously.

I could use some guidance for writing tests.

### DIFF
--- a/src/pegen/python_generator.py
+++ b/src/pegen/python_generator.py
@@ -237,8 +237,8 @@ class PythonParserGenerator(ParserGenerator, GrammarVisitor):
 
         self.print()
         with self.indent():
-            self.print(f"KEYWORDS = {tuple(self.callmakervisitor.keywords)}")
-            self.print(f"SOFT_KEYWORDS = {tuple(self.callmakervisitor.soft_keywords)}")
+            self.print(f"KEYWORDS = {tuple(sorted(self.callmakervisitor.keywords))}")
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
I am not sure either how to best test this feature. And actually currently no test validate the content of the KEYWORDS and SOFT_KEYWORDS attributes. So you could add a test under test_pegen.py generating a parser from a small grammar containing both hard and soft keywords and testing that the output is as expected. The other tests there should illustrate how to easily generate the parser class. Let me know if you need further guidance.

edemaine:
Thanks for the guidance! I've added a test that seems to do the job.

lysnikolaou (approved):
LGTM! Thanks @edemaine for the PR!

pablogsal:
Thanks, @edemaine! 🎉

Issue #47, opened 2021-10-03
python.gram: Incorrect column numbers in SyntaxError


edemaine:
Python's `tokenize.TokenInfo` structure has 0-indexed column numbers, but `SyntaxError` wants 1-indexed column numbers. See [this discussion](https://bugs.python.org/issue40546).

This code from `python.gram` is thus incorrect:

https://github.com/we-like-parsers/pegen/blob/d1300381b1de5c5534bfbc148fefdfe65c644ae4/data/python.gram#L306-L309

On the other hand, this code from `parser.py` is correct:

https://github.com/we-like-parsers/pegen/blob/d1300381b1de5c5534bfbc148fefdfe65c644ae4/src/pegen/parser.py#L262

Could be fixed as part of #41.

MatthieuDartiailh:
Thanks for the pointer @edemaine . I may not have time to work on #41 in the next couple of weeks (Python 3.10 final release means a bunch of CIs updates and releases for me) so feel free to make a PR against my branch.

edemaine:
Sorry for dropping the ball here.  I tried adding tests against actual Python, and found lots of matches, but also many discrepancies.  Here's an example on Python 3.10.0:

```py
>>> {**c, a: *b}
  File "<stdin>", line 1
    {**c, a: *b}
             ^^
SyntaxError: cannot use a starred expression in a dictionary value
```

By contrast, pegen thinks the error involves one more character:

```py
  File "<unknown>", line 1
    {**c, a: *b}
             ^^^
SyntaxError: cannot use a starred expression in a dictionary value
```

There's also discrepancies between Python versions.  For example, here's a Python 3.9.7 error:

```py
>>> {a: 1, c: 2, b}
  File "<stdin>", line 1
    {a: 1, c: 2, b}
                  ^
SyntaxError: invalid syntax
```

And here's the corresponding Python 3.10.0 error:

```py
>>> {a: 1, c: 2, b}
  File "<stdin>", line 1
    {a: 1, c: 2, b}
                 ^
SyntaxError: ':' expected after dictionary key
```

And here's pegen's error:

```py
  File "<unknown>", line 1
    {a: 1, b}
          ^^
SyntaxError: ':' expected after dictionary key
```

Overall, there are 68 discrepancies from Python 3.10 (and 72 from Python 3.9).

[...]

PR #48, opened 2021-10-19
python.gram: SyntaxError column numbers to match cpython 3.10

edemaine:
As suggested by @MatthieuDartiailh in [this comment](https://github.com/we-like-parsers/pegen/issues/47#issuecomment-933217996), this is a completion of #41 (which includes @MatthieuDartiailh original commits) that includes all the row/column tests, as well as a fix to my issue #47.

### DIFF
--- a/data/python.gram
+++ b/data/python.gram
@@ -1580,10 +1580,10 @@ invalid_arguments[NoReturn]:
             "invalid syntax. Maybe you meant '==' or ':=' instead of '='?", a, b
         )
      }
-    | a=args for_if_clauses {
-        self.store_syntax_error_starting_from(
+    | a=args b=for_if_clauses {
+        self.store_syntax_error_known_range(
             "Generator expression must be parenthesized",
-            a[1][-1] if a[1] else a[0][-1]
+            a[1][-1] if a[1] else a[0][-1], b
         )
      }
     | args ',' a=expression b=for_if_clauses {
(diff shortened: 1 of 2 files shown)


cfbolz:
hi everyone! I am considering to improve a few things in python.gram but don't want to tread on anybody's else's effort. Should I help out with the failing tests here?

(I was wondering whether the reason for the test failures is that `python_generator.py` does not seem to special-case the `invalid_`-prefixed rules in the same way that CPython's `c_generator.py` does it?)

edemaine:
I'd definitely appreciate the help with fixing these tests! (Though not sure whether @MatthieuDartiailh is currently working on it.) And we should make the comparison to Python only run on 3.10 (and maybe even a specific minor version?).

MatthieuDartiailh:
I am not working on this at the moment no.

pablogsal:
> should probably be updated to use 3.10.0.

Done in https://github.com/we-like-parsers/pegen/pull/50

MatthieuDartiailh:
If nobody is working on this I will try to pick this up again soon now that my proof of concept for using pegen in a project I maintain is ready. (https://github.com/nucleic/enaml/pull/474).

I will probably first look into using the two passes error reporting mechanism used in the C parser.

cfbolz:
I didn't get around to it sorry :-(

MatthieuDartiailh:
With #41 merged I believe everything this addressed has been fixed so closing. Feel free to reopen if the need arises.

PR #50, opened 2021-11-10
Update the CI to use Python 3.10

pablogsal:


edemaine:
Did you want to keep Pythons 3.8 and 3.9 in the matrix, to make sure the parser builds/runs there too? The tests in #48 can be restricted to only run in 3.10.0.

PR #56, opened 2022-02-11
Add missing tox test requirement

SeanDS:
`tox` is missing from the test dependencies. This PR adds it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
     extras_require={
         'docs': ['sphinx', 'sphinx-copybutton', 'furo'],
         'lint': ['black', 'flake8', 'mypy'],
-        'test': ['pytest', 'pytest-cov'],
+        'test': ['pytest', 'pytest-cov', 'tox'],
     },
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -33,7 +33,7 @@
     extras_require={
         'docs': ['sphinx', 'sphinx-copybutton', 'furo'],
         'lint': ['black', 'flake8', 'mypy'],
-        'test': ['pytest', 'pytest-cov'],
+        'test': ['pytest', 'pytest-cov', 'tox'],
```
pablogsal, line 36: Normally tox is not needed to run the test suite but to orchestrate it. This will also make tox install itself when running the test suite.

pablogsal:
I'm closing this as `tox` is not really a test requirement that needs to be installed to run the test suite.

SeanDS:
Fair enough, though I did have some error when running `pytest` after a fresh install (can't remember what) and got it to work by installing and running `tox`.

Issue #53, opened 2022-01-24
Remove hard dependencies


MatthieuDartiailh:
Currently we have hard dependencies on psutil, flask and flaks-wtf which seems completely overkill since one can perfectly use pegen without any of those.

MatthieuDartiailh:
Looking in details at the use of each dependency:
- psutil is used in tests/utils and is already behind a try, except block. I believe it can simply be removed. We could add a 'memory' extra_requires though.
- flask and flask-wtf are used in the web.py module which provides a web based playground from reading quickly. Do we want to keep it ? @pablogsal  If we want to keep this making the dependency optional should not be a problem and allowing to access the module as an executable would make sense (pegen-web-playground ?). It may also duplicate some utilities but I did not check in details yet.

pablogsal:
> Looking in details at the use of each dependency:
> - psutil is used in tests/utils and is already behind a try, except block. I believe it can simply be removed. We could add a 'memory' extra_requires though.
> - flask and flask-wtf are used in the web.py module which provides a web based playground from reading quickly. Do we want to keep it ? @pablogsal  If we want to keep this making the dependency optional should not be a problem and allowing to access the module as an executable would make sense (pegen-web-playground ?). It may also duplicate some utilities but I did not check in details yet.
> 

We can make flask and flask-wtf optional, as is not required for the core functionality of pegen. We can make this install under `pegen[web]` for example. Optionally we could even move this to a separate package, but they may be too much. What are your thoughts?

MatthieuDartiailh:
I am fine having them under pegen[web] but I will audit the different helpers found in web.py. At first glance they look a lot like some testing helpers.

MatthieuDartiailh:
True. If nobody beats me to it I will look into it but probably after dealing with improving the error reporting of the parser.

lysnikolaou:
@MatthieuDartiailh Can we close this now?

MatthieuDartiailh:
I believe so yes. Sorry I forgot to close it.

PR #58, opened 2022-03-20
Remove hard dependencies

smason:
just tried using this package and was surprised by the dependency on flask

first patch seems inline with suggestions from #53, while the second is just a convenience thing but seemed to go along with editing `setup.py`


smason:
#59 looks much more comprehensive!

PR #59, opened 2022-03-20
Clean utils and install

MatthieuDartiailh:
This move the re-usable bits from tests.utils to pegen.utils and remove the duplication of a large number of them from web.py. It also clean the dependencies in setup.py (like #58 but does not introduce the script).


lysnikolaou (approved):
LGTM!

PR #65, opened 2022-04-18
Allow using an alternative tokenizer

MatthieuDartiailh:
I am not sure this approach is the nicest and it looks a bit brittle in places but it would allow to use alternate tokenizers with a well defined interface which would be nice.

Comments welcome !


src/pegen/tokenizer.py
```diff
@@ -1,27 +1,121 @@
...
-    This is pretty tied to Python's syntax.
     """
 
-    _tokens: List[tokenize.TokenInfo]
+    _tokens: List[TokenInfo[TokenType]]
 
     def __init__(
-        self, tokengen: Iterator[tokenize.TokenInfo], *, path: str = "", verbose: bool = False
+        self,
+        tokengen: Iterator[TokenInfo[TokenType]],
```
funkyfuture, line 112: why not name it `token_generator`?

src/pegen/tokenizer.py
```diff
@@ -1,27 +1,121 @@
...
-class Tokenizer:
-    """Caching wrapper for the tokenize module.
+    type: TokenType
+    string: str
+    start: tuple[int, int]
+    end: tuple[int, int]
+    line: str
+
+
+class TokenTypes(Generic[TokenType]):
```
funkyfuture, line 23: the annotations of abstract methods don't result in an exception when they are not implemented as the class doesn't derive from `ABC`. if that is intended, it should be mentioned, or the methods can raise a `NotImplementedError`.

src/pegen/tokenizer.py
MatthieuDartiailh: This predate this work so I did not feel like changing it.

src/pegen/tokenizer.py
MatthieuDartiailh: Fixed

MatthieuDartiailh:
Rebased but will #71 to pass on 3.10

src/pegen/tokenizer.py
MatthieuDartiailh: Fixed in latest commit

MatthieuDartiailh:
Rebased again so that CIs should pass @pablogsal once the 3.11beta5 is live could you review this ?

MatthieuDartiailh:
@pablogsal I resolved the conflict and hopefully fixed the 3.8 failure. Could you please review ?

0dminnimda:
It would be nice to finally merge this

PR #68, opened 2022-07-03
Stop creating unused variables

0dminnimda:


0dminnimda:
Is it reasonable first try to merge it to cpython?

pablogsal:
> Is it reasonable to first try to merge it to cpython?

Generally we do it the other way: we merge stuff here and I sync them from time to time in batches.

0dminnimda:
Ok, I see.
Also can you ask for the launch of the workflow?

0dminnimda:
Probably a good time to say, that two tests are failing on the `main`. The same tests are failing here. Although I think fixing those tests is out of the scope of this pr, unfortunately

MatthieuDartiailh:
Getting #66 in first would make sense to get CIs to test on more versions.

0dminnimda:
Does #66 fixes broken tests?

0dminnimda:
> Generally we do it the other way: we merge stuff here and I sync them from time to time in batches

Well, then what's about distributing credit/blame for the changes? Will those changes be cherry-picked into cpython?

pablogsal:
> > Generally we do it the other way: we merge stuff here and I sync them from time to time in batches
> 
> Well, then what's about distributing credit/blame for the changes? Will those changes be cherry-picked into cpython?

No, they will be directly included and blame/credit is lost.

0dminnimda:
> No, they will be directly included in one batch and blame/credit is lost.

Aww, isn't this a reason to make a direct pr to cpython?

Well, not like it's *that* important, but it would be nice if the credit/blame was preserved!

MatthieuDartiailh:
@0dminnimda Can you rebase on main so that we get an extended test run ?

0dminnimda:
Oh yeah, workflow activation, what a lovely github feature ♥

gvanrossum:
Bye, I am leaving this repo now.

0dminnimda:
oh .. bye

<br>
<br>
<br>

well, we need to get back to the work

3.10 didn't pass, it seems like versoin of python that tox runs is somehow different from the other ones
Because if I try replicate this I see a string without `'`

I don't quite understand why it happens

MatthieuDartiailh:
#71 will fix the failure on 3.10 (3.10.5 fixed an issue for which we had a workaround)

0dminnimda:
Closing and opening to triger the ci

0dminnimda:
oh yeah, forgot to merge

0dminnimda:
Is there any additional comments or this can be merged?

src/pegen/python_generator.py
```diff
@@ -198,6 +198,25 @@ def visit_Forced(self, node: Forced) -> Tuple[str, str]:
             )
 
 
+Used = Set[str]
```
MatthieuDartiailh, line 201: I tend to find that kind of aliasing to be detrimental to the readability personally (in particular for such a simple alias). But I will let @pablogsal comment on this.

src/pegen/grammar_parser.py
```diff
@@ -248,21 +236,16 @@ def named_item(self) -> Optional[NamedItem]:
         if (
             (name := self.name())
             and (annotation := self.annotation())
-            and (literal := self.expect("="))
-            and (cut := True)
+            and (self.expect("="))
+            and (True)
```
MatthieuDartiailh, line 240: This looks wrong to me. Updating cut has side effects (l 268) and they are not taken into account.

MatthieuDartiailh (changes requested):
I found one concerning issue regarding cut (and I find even more concerning that no test found the issue). I also left a couple of comment/question.

src/pegen/grammar_parser.py
0dminnimda: oh yeah, I didn't handle cut properly

src/pegen/python_generator.py
```diff
@@ -319,6 +346,18 @@ def visit_Rhs(self, node: Rhs, is_loop: bool = False, is_gather: bool = False) -
     def visit_Alt(self, node: Alt, is_loop: bool, is_gather: bool) -> None:
         has_cut = any(isinstance(item.item, Cut) for item in node.items)
         has_invalid = self.invalidvisitor.visit(node)
+
+        used = None
+        if node.action:
+            # Those should be kw arguments
```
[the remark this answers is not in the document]
0dminnimda: Well, it raises another issue, what if the LOCATIONS formatting does not use keyword args as are now set bu the default. In this case this will break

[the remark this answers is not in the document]
0dminnimda: I'm gonna try to make formating apply before the usednamesvisitor

0dminnimda:
Maybe @pablogsal have other comments?

src/pegen/python_generator.py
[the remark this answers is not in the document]
0dminnimda: Also, @MatthieuDartiailh do you have any comments regarding new implementation? 
If no, you can resolve this convo.

src/pegen/python_generator.py
```diff
@@ -316,9 +342,62 @@ def visit_Rhs(self, node: Rhs, is_loop: bool = False, is_gather: bool = False) -
...
+                action = action.replace("LOCATIONS", self.location_formatting)
+            if "UNREACHABLE" in action:
+                unreachable = True
+                action = action.replace("UNREACHABLE", self.unreachable_formatting)
+
+        used = None
+        if action:
+            used = self.usednamesvisitor.visit(ast.parse(action))
+            if has_cut:
+                used.add("cut")
```
MatthieuDartiailh, line 399: No need for the 2 ifs

```suggestion
        locations = False
        unreachable = False
        used = None
        if action:
            if "LOCATIONS" in action:
                locations = True
                action = action.replace("LOCATIONS", self.location_formatting)
            if "UNREACHABLE" in action:
                unreachable = True
                action = action.replace("UNREACHABLE", self.unreachable_formatting)

            used = self.usednamesvisitor.visit(ast.parse(action))
            if has_cut:
                used.add("cut")
```

src/pegen/python_generator.py
```diff
@@ -299,12 +316,21 @@ def visit_Rule(self, node: Rule) -> None:
...
+        elif node.name:
             name = node.name
+
+        if used is not None and name not in used:
+            name = None
+
         if not name:
-            self.print(call)
+            # Parentheses are needed because the trailing comma may appear :>
+            self.print(f"({call})")
```
MatthieuDartiailh, line 333: Could you just point me to a concrete example of this issue ?

MatthieuDartiailh (changes requested):
I left a minor suggestions for improvement and a question otherwise LGTM.

src/pegen/python_generator.py
0dminnimda: oh, yeah! right 😁

0dminnimda: Well, actually .. I think the original version is more readable. Not like the second if will lead to critical performance hit.
Here we have two if's for different logic, easy to follow and change

0dminnimda: Idk, I'll left it for the final review by the pablogsal

src/pegen/python_generator.py
0dminnimda: Woulld love to!
You cannot find it in the metaparser, but in the python parser.

0dminnimda: For example:
```python
@memoize
    def file(self) -> Optional[ast . Module]:
        # file: statements? $
        mark = self._mark()
        if (
            (a := self.statements(),)
            and
            (self.expect('ENDMARKER'))
        ):
            return ast . Module ( body = a or [] , type_ignores = [] );
        self._reset(mark)
        return None;
```

0dminnimda: As it seems all of the rules like `X?` will result in this

0dminnimda: And `[X]` will also result in this

0dminnimda: It seems logical, optional rule cannot be incorrect as if there's nothing found it's still fine!

src/pegen/python_generator.py
MatthieuDartiailh: IMO having a single if with a comment before the inner ifs and before the systematic part would be just as readable without duplicating the if test.

src/pegen/python_generator.py
MatthieuDartiailh: I need to look at how things work currently and why this is now needed.

MatthieuDartiailh: So I checked and in the case of the file rule the parentheses are already there for the case of statements so I believe we do not need this change or am I missing something ?

0dminnimda: parentheses are there because the name is used so
`self.print(f"({name} := {call})")` is responsible
But if we don't use the name we still need to have parentheses to handle the comma
`(X := y,) == (y,)`

0dminnimda: Generate a python parser with this branch and search for `,)` you'll see the examples of both of those cases

MatthieuDartiailh: Now I get it. I just wished we could do it in such a way we do not generate extra parenthesis but I guess black can clean up for us. Thanks for bearing with me.

0dminnimda: You're welcome!
Why do you not like generating parentheses in there?

MatthieuDartiailh: The ones around the endmarker in the example above look funny to me but it is not a big deal.

0dminnimda:
@MatthieuDartiailh can you allow the workflow?

MatthieuDartiailh:
Done

MatthieuDartiailh:
@pablogsal would you have time to look at this ? or do you want to defer the final review to me maybe (since you seem quite busy ATM).

pablogsal:
> @pablogsal would you have time to look at this ? or do you want to defer the final review to me maybe (since you seem quite busy ATM).

I apologize for the late reply @MatthieuDartiailh. Unfortunately the release of 3.11 is being ridiculously challenging and I'm expending all my OSS time on that :(

Please, go ahead and merge it if you feel confident. I trust your criteria :+1:

src/pegen/python_generator.py
MatthieuDartiailh: If you can address this comment @0dminnimda I will merge this PR. This the last blocker on my side and I got the green light from @pablogsal

0dminnimda: It's not that big of a deal

MatthieuDartiailh: I will add some comments to the different branches and merge when I get a chance.

MatthieuDartiailh:
I wanted to add a couple small comments but apparently I cannot push to your repo. S I will merge and add the comment directly on main.

0dminnimda:
Weird usually there's no problem in that
Maybe i accidentally checked out Allow edits by maintainers..

PR #70, opened 2022-07-11
Generate pxd

### LINKED ISSUE #69: [Feature] Optionally create `.pxd` files
If the creation of a parser using c was prohibited, because the original cpython pegen used too much internal functionality, then we can at least make python parser faster with cython.

Since pegen knows all the variables and types that users provide, it can use this information to create a `.pxd` file. This file will not interfere with pure python parser work, but it can be used by cython to create a faster compiled version of the parser.

I will implement this functionality anyway, but I wonder should I make it into PR and try to merge it here? Is this feature needed here?

0dminnimda:
Closes #69

<details>
  <summary></summary>

  nice, sorry ;]

</details>


MatthieuDartiailh:
Could we get some benchmarks ? Given the workflow of pegen it is not obvious to me how much cython will be able to optimize things. In particular since we would have to drop the walrus operator which makes the generated parser more readable IMO.

MatthieuDartiailh:
Also if we can wait for Cython 3.0 we will get walrus support (see https://github.com/cython/cython/issues/2636). However I have no idea regarding the timeline for its release.

0dminnimda:
> Also if we can wait for Cython 3.0 we will get walrus support (see [cython/cython#2636](https://github.com/cython/cython/issues/2636)). However I have no idea regarding the timeline for its release.

It's very unlikely that it'll be out in 2022
There are not a lot of people working on cython, so unfortunately 3.0 will not be out soon

0dminnimda:
> Could we get some benchmarks ?

Of course

Also regarding drop of the walrus. I'll merge `no_walrus` branch here because changes will not work without it.
And if we'll be fine with that change I will open a pr for this branch, so it could be merged first

jnoortheen:
fyi, walrus operator support is now released with 3.0a11 version https://github.com/cython/cython/blob/master/CHANGES.rst#300-alpha-11-2022-07-31

0dminnimda:
@jnoortheen oh well, now there are better news, the [beta 3.0.0b1](https://github.com/cython/cython/releases/tag/3.0.0b1) was finally released!

I actually made some good progress on this and stuck trying to make memorization decorators to be optimized as well
if I remember correctly I did not push changes that I was not satisfied with / that didn't really work

PR #71, opened 2022-07-12
Cis: use newer actions (setup-python, checkout, cache)

MatthieuDartiailh:
And fix the one broken test on 3.10.5. Setuptools-scm complains but appear to work so I won't touch the fetch.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
     - uses: actions/checkout@master
     - name: Set up Python 3.8
-      uses: actions/setup-python@v2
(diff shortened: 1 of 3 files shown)


pablogsal:
What's the complain from setuptols-scm?

MatthieuDartiailh:
Setuptools complain the repository is shallow but since we have the tags things kind of work. To avoid it we would need to specify a fetch depth of 0.

MatthieuDartiailh:
@pablogsal anything you would like to see before merging ?

pablogsal:
> @pablogsal anything you would like to see before merging ?

No, everything looks good. Great job!

PR #73, opened 2022-07-26
Add ability to run each test separately, thus easier debugging

0dminnimda:
Coincidently format imports

### DIFF
--- a/tests/python_parser/test_ast_parsing.py
+++ b/tests/python_parser/test_ast_parsing.py
@@ -94,3 +94,7 @@ def test_parser(python_parse_file, python_parse_str, filename):
     p = ast.dump(python_parse_file(path), **kwargs)
     diff = "\n".join(difflib.unified_diff(o.split("\n"), p.split("\n"), "cpython", "python-pegen"))
     assert not diff
+
+
+if __name__ == "__main__":
+    pytest.main([__file__])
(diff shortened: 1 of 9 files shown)


MatthieuDartiailh:
This can already be achieved by specifying which file to run when invoking pytest so I do not see the value honestly. The formatting however is always nice (in a different PR) but we should also figure out how to have the CIs ensure it. I believe it required to touch too much the tox config last time I checked but I am not a tox expert by any means.

Feel free to re-open if I missed something.

0dminnimda:
> This can already be achieved by specifying which file to run when invoking pytest

So, if I'm corrct, you propose to use the same code (in the other file I guess) but just specify the file to debug.
Well, it'll work but it's not really convinient, opposed to just running the file in the debug mode and not worrying about paths and stuff ¯\_(ツ)_/¯.

At least I can say that I find myself editing tests that don't have those lines and it's not convinient to add them in each branch

Of coure if you thinkg it don't fit the repo, you can left this closed, but it defenetly impoves the debug process and is a more friendly way

MatthieuDartiailh:
I guess the main issue is that I do not understand your workflow. Since the testsuite uses pytest I run it using `pytest tests` . If I want to run a single file I just specify it and to drop to pdb on a failed test I use the `--pdb` option. As a consequence I do not see the need to add pytest.main to every file (I actually never used it).

0dminnimda:
Oh, well I use ide for debugging

PR #77, opened 2022-09-25
Fix empty last line returned by tokenizer

### LINKED ISSUE #76: Tokenizer returns empty string for the last line
Hi 👋  I'm using pegen for parsing a Python-like language, encountered something unexpected, and wondering if anyone can take a look.

## Problem

It seems that `pegen.tokenizer.Tokenizer` stores the empty string for the last line of the source if the source does not end with a NEWLINE.

```python
import io
import tokenize

from pegen.tokenizer import Tokenizer

source = "line 1\nline 2\nline 3"

print(source)

tok_stream = tokenize.generate_tokens(io.StringIO(source).readline)
tokenizer = Tokenizer(tok_stream)

# load all tokens
while True:
    tok = tokenizer.getnext()
    print(tok)
    if tok.type == tokenize.ENDMARKER:
        break

print(tokenizer._lines)

print("last line: ", tokenizer.get_lines([3]))
```

I expect the last line to print `["line 3"]` but it prints `['']`.

After some investigation, I found that the empty line is coming from NEWLINE tokens generated by Python's tokenize module.

If the source does not end with a NEWLINE, tokenize injects one. For example, tokenizing the above input would produce

```
TokenInfo(type=1 (NAME), string='line', start=(1, 0), end=(1, 4), line='line 1\n')
TokenInfo(type=2 (NUMBER), string='1', start=(1, 5), end=(1, 6), line='line 1\n')
TokenInfo(type=4 (NEWLINE), string='\n', start=(1, 6), end=(1, 7), line='line 1\n')
TokenInfo(type=1 (NAME), string='line', start=(2, 0), end=(2, 4), line='line 2\n')
TokenInfo(type=2 (NUMBER), string='2', start=(2, 5), end=(2, 6), line='line 2\n')
TokenInfo(type=4 (NEWLINE), string='\n', start=(2, 6), end=(2, 7), line='line 2\n')
TokenInfo(type=1 (NAME), string='line', start=(3, 0), end=(3, 4), line='line 3')
TokenInfo(type=2 (NUMBER), string='3', start=(3, 5), end=(3, 6), line='line 3')
TokenInfo(type=4 (NEWLINE), string='', start=(3, 6), end=(3, 7), line='') 👈
TokenInfo(type=0 (ENDMARKER), string='', start=(4, 0), end=(4, 0), line='')
```

The second-to-last token does not exist in the original source but was added by the tokenizer. The problem is that this injected token does not have `line` (which on its own is reasonable).

https://github.com/we-like-parsers/pegen/blob/fab0c5b012836fcba07c1c1c828874745c8a4bfd/src/pegen/tokenizer.py#L59

Pegen's tokenizer uses this to cache the text on that line. So after execution, `tokenizer._lines` is set to `{1: 'line 1\n', 2: 'line 2\n', 3: '', 4: ''}`.

## Possible Fixes

If this is an unintentional behavior, a possible fix I can think of is to update `_lines` only if the same line has not been set.

I've added a test case, implemented this change, and it seems to be working.

https://github.com/we-like-parsers/pegen/compare/main...shumbo:pegen:fix-last-line?expand=1

Is there anything else to consider, especially around the condition of the if statement? I can create a PR if this may benefit someone else.

shumbo:
Fixes #76 

This PR fixes an issue where `tokenizer` returns the empty string for the last line if no NEWLINE exists at the end of the source.

Python's `tokenize` module injects a NEWLINE if the source does not end with one. The injected token has the empty string for its `line` but Pegen's tokenizer remembers that empty string as the content of the last line of the source.

Since source does not change over tokenization, once `self._lines[i]` is set for some `i`, it does not need to be set again with later tokens. I added a check to only update `self._lines` only if `tok.start[0]` is not set. With this change, `NEWLINE` injected by the `tokenize` module will not override `self._lines` with the empty string.

I also added a test case for this change.

### DIFF
--- a/releasenotes.rst
+++ b/releasenotes.rst
@@ -11,6 +11,7 @@ unreleased: Version 0.2.0
 - add documentation PR #43 #52
 - sort KEYWORDS to make output deterministic PR #44
 - update grammar_grapher with the new forced (&&) directive PR #57
+- fixed bug where tokenizer reported the last line of source as empty #77
 
 09/06/2021: Version 0.1.0
 -------------------------
--- a/src/pegen/tokenizer.py
+++ b/src/pegen/tokenizer.py
@@ -55,7 +55,7 @@ class Tokenizer:
             ):
                 continue
             self._tokens.append(tok)
-            if not self._path:
+            if not self._path and tok.start[0] not in self._lines:
                 self._lines[tok.start[0]] = tok.line
         return self._tokens[self._index]
 
--- a/tests/test_tokenizer.py
+++ b/tests/test_tokenizer.py
@@ -1,5 +1,5 @@
 import io
-from tokenize import NEWLINE, NUMBER, TokenInfo, generate_tokens
+from tokenize import NEWLINE, NUMBER, ENDMARKER, TokenInfo, generate_tokens
 
 from pegen.tokenizer import Tokenizer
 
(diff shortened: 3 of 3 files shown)


MatthieuDartiailh (commented):
Looks good to me. Could you add an entry in the changelog ?

shumbo:
@MatthieuDartiailh Updated the changelog in 487ebe07d7a02b4c450806f9217125ac11ddd083

PR #84, opened 2023-02-25
Rewrite parser generator 

heckad:
What does this pull request do?

This pull request changes what we consider successful parsing. Now we have the `not_parsed` marker (the name can be changed) which mark that rule can't parse input. This solves the problem with empty collections being interpreted as parsing failure and also allows you to drag None up. Unfortunately this is a very big change. Parsers generated by the previous version will not work. However, this solves all problems.


src/pegen/parser.py
```diff
@@ -15,6 +16,14 @@
 F = TypeVar("F", bound=Callable[..., Any])
 
 
+@final
+class NotParsedMarker(Enum):
+    marker = 0
+
+
+not_parsed = NotParsedMarker.marker
```
0dminnimda, line 24: You can just use `object()` and it'll give the same effect
Also `@final` doesn't do a lot, It's better to just use `Literal[not_parsed]` for annotations

data/python.gram
```diff
@@ -488,7 +488,7 @@ statement_newline[list]:
     | a=compound_stmt NEWLINE { [a] }
     | simple_stmts
     | NEWLINE { [ast.Pass(LOCATIONS)] }
-    | ENDMARKER { None }
+    | ENDMARKER { [] }
```
0dminnimda, line 491: Why this change was made?

src/pegen/parser.py
```diff
@@ -263,16 +272,19 @@ def positive_lookahead(self, func: Callable[..., T], *args: object) -> T:
         self._reset(mark)
         return ok
 
-    def negative_lookahead(self, func: Callable[..., object], *args: object) -> bool:
+    def negative_lookahead(self, func: Callable[..., object], *args: object) -> Union[Literal[True], NotParsedMarker]:
         mark = self._mark()
         ok = func(*args)
         self._reset(mark)
-        return not ok
+        return not_parsed if ok is not not_parsed else True
```
0dminnimda, line 279: bool check is more than enough here and in positive_lookahead

0dminnimda (commented):
Good attempt 

Also, if you are using someone else's ideas or code it's nice to mention them

data/python.gram
heckad: None is now a valid result but we expect a list to be returned there.

src/pegen/parser.py
heckad: No, `False` now a correct parsing result. For mark incorrect parsing result we always should use `not_parsed`.

heckad:
> Also, if you are using someone else's ideas or code it's nice to mention them

I don't use any another idea. I just improved my last [attempt](https://github.com/we-like-parsers/pegen/pull/83). Now `None`, `False` and any other object are correct parsing results.

heckad:
@MatthieuDartiailh, please look at pull request.

PR #85, opened 2023-03-23
Merge pegen from cpython

phorward:
This is a merge of the current pegen state (commit hash 87be8d95228ee95de9045cf2952311d20dc5de45) from cpython into pegen.

I've also made some code restructuring:
1. The nullable detection went into grammar.py, and
2. the classes in parser_generator.py where re-arranged and make use of the grammar rather than the rules.

I'm planning to do further restructuring soon. `pegen` should become the base for a project idea I would like to start, therefore its modules and classes will be cleaned-up and re-arranged.


MatthieuDartiailh:
Given the currently low level of availability from the maintainers, could you try to focus on a single change at a time ?

phorward:
> Given the currently low level of availability from the maintainers, could you try to focus on a single change at a time ?

Hello @MatthieuDartiailh, okay I will try to do so, and close this PR for now.
In general, are there plans for the future of pegen? Maybe we can discuss this also by email or similar.

Issue #93, opened 2023-08-28
Pegen Crashes on f-strings in 3.12


Eric-Vin:
Pegen seems to crash when trying to parse f-strings in Python 3.12. I'm guessing this is because Pegen is invoking the 3.12 tokenizer that now uses several new tokens for f-strings, not just `STRING`. 

To reproduce I installed Pegen in 3.12 and ran `python -m pegen data/python.gram`, which resulted in the following error:

```
  File "<unknown>", line 1690
    self.raise_syntax_error_known_range(f"cannot assign to {a.string}", a, b)
                                        ^
SyntaxError: data/python.gram
```

MatthieuDartiailh:
The Python parser needs to be updated to include the changes to the grammar in 3.12 indeed. Would you be interested in looking at the required changes ?

dfremont:
Hello Matthieu, I've made a draft PR (#94) adding support for the 3.12 tokenizer and fixing the problem Eric mentioned above -- if you could take a look that would be great. Please note I haven't attempted to update the Python grammar for 3.12, since that's an orthogonal issue.

lysnikolaou:
Closed in #94. Thanks @Eric-Vin for the report and @dfremont for the fix.

PR #94, opened 2023-09-01
Support Python 3.12 f-strings in grammar actions

dfremont:
As pointed out in #93, in Python 3.12 the PEP 701 changes to f-string tokenization cause Pegen to fail on grammars which contain f-strings in their actions (such as the Python grammar). This PR adds support for the new tokens to Pegen and changes the metagrammar to restore f-string support under 3.12 (while preserving support under older versions of Python).

Two things this PR does not currently do:
- Fix broken tests under 3.12. I'm getting 28 failures currently, but as far as I can tell they don't have to do with the changes in this PR and are rather caused by other differences in the 3.12 tokenizer/parser. I can try to fix these later, but any help would be appreciated!
- Update the example Python grammar (`data/python.gram`) for 3.12.

### DIFF
--- a/src/pegen/metagrammar.gram
+++ b/src/pegen/metagrammar.gram
@@ -126,6 +126,13 @@ target_atom[str]:
     | NAME { name.string }
     | NUMBER { number.string }
     | STRING { string.string }
+    | l=FSTRING_START m=target_fstring_middle* r=FSTRING_END { l.string + ''.join(m) + r.string }
     | "?" { "?" }
     | ":" { ":" }
     | !"}" !"]" OP { op.string }
+
+target_fstring_middle[str]:
+    | FSTRING_MIDDLE { fstring_middle.string }
+    | "{" { "{" }
+    | "}" { "}" }
+    | target_atom { target_atom }
--- a/src/pegen/metagrammar.gram
+++ b/src/pegen/metagrammar.gram
@@ -126,7 +126,7 @@ target_atom[str]:
     | NAME { name.string }
     | NUMBER { number.string }
     | STRING { string.string }
-    | l=FSTRING_START m=target_fstring_middle* r=FSTRING_END { l.string + ''.join(m) + r.string }
+    | l=FSTRING_START m=target_fstring_middle* r=FSTRING_END { l.string + "".join(m) + r.string }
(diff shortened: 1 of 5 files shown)


MatthieuDartiailh (commented):
Looks like a good start. Could you please run black on the generated parser ?

dfremont:
Ah, that explains why it looked funny after I regenerated it!

src/pegen/metagrammar.gram
```diff
@@ -126,6 +126,13 @@ target_atom[str]:
     | NAME { name.string }
     | NUMBER { number.string }
     | STRING { string.string }
+    | l=FSTRING_START m=target_fstring_middle* r=FSTRING_END { l.string + ''.join(m) + r.string }
     | "?" { "?" }
     | ":" { ":" }
     | !"}" !"]" OP { op.string }
+
+target_fstring_middle[str]:
```
pablogsal, line 134: This looks odd, the metragrammar should not need changes

pablogsal (approved):
This PR should modify the python grammar, not the `metagrammar.gram` because f-strings do not exist in grammar files.

src/pegen/metagrammar.gram
pablogsal: Ah, wait, is this for the action part of the grammar files?

dfremont:
Thanks for looking this over, Pablo. I'll mark this ready for review in case you and Matthieu want to merge it without fixing the 3.12 test failures.

MatthieuDartiailh:
The linting failures should be fixed before merging.

dfremont:
Fixed!

MatthieuDartiailh:
Thanks. The regeneration test still seems unhappy though.

dfremont:
Whoops, I should have run `tox`. The problem is that the `PythonParserGenerator` gets its list of tokens from the `token` package, which prior to 3.12 doesn't include `FSTRING_START` etc. I've added them manually.

phorward (approved):

pablogsal:
Thanks for the PR @MatthieuDartiailh

pablogsal:
Thanks for the PR @dfremont  !

PR #95, opened 2023-09-11
Support Python 3.11 and 3.12

MatthieuDartiailh:


phorward:
Hello @MatthieuDartiailh, thanks for starting on this issue. I've already updated pegen in my fork [unipeg](https://github.com/phorward/unipeg) to Python 3.10 and Python 3.11, maybe you can use parts of it here as well.
Generally, my advise is to drop support for Python < 3.11, and stay up-to-date.

Also, I would like to turn pegen (or unipeg) into a drop-in replacement of the [peg](https://github.com/python/cpython/tree/main/Tools/peg_generator/pegen) provided with CPython. I know this isn't the right place for discussions, but we should start to discuss about it and bring this project further.

MatthieuDartiailh:
All ast parsing tests are now passing on Python 3.12 and I added a bunch from CPython for tryexcept group, type params and f-strings.

I still need to fix syntax error related tests in a manner compatible both with 3.12 and earlier version. 

I would welcome an early review.

MatthieuDartiailh:
Tests for Python 3.11 and 3.12 specific syntax error messages are missing but everything else is good and those tests could be added in a follow up PR.

A review would be welcome.

MatthieuDartiailh:
I added some extra tests. I believe this is ready to go from my end.

MatthieuDartiailh:
Thanks for checking the PR @phorward

lysnikolaou:
@MatthieuDartiailh Thanks for working on this! I'll probably try to review this tomorrow.

pablogsal (approved):
I have a couple of passes and I couldn't find anything obviously wrong. I am a bit short on cycles so I couldn't gave a through review, but given that @phorward took a look I think we can just land this and fix anything that doesn't work afterwards.

This is fantastic work @MatthieuDartiailh, thanks a lot for doing this ✨

MatthieuDartiailh:
@lysnikolaou friendly ping

lysnikolaou (approved):
I had a quick look over the weekend and couldn't find anything very obviously off. Let's merge and fix stuff as they come.

Thanks for doing this @MatthieuDartiailh! 🚀

PR #96, opened 2023-10-31
Added fstring Test Case and Potential Fix

Eric-Vin:
There seems to be an fstring case that is incorrect with the current Python grammar. I've added a case to fstrings.py to highlight the behavior. I think the issue is an unnecessary encode in the `fstring_format_spec` rule, so I removed that, but I'm not sure if that has other implications so I would appreciate any feedback.

### DIFF
--- a/data/python.gram
+++ b/data/python.gram
@@ -1696,7 +1696,7 @@ fstring_full_format_spec:
         )
      }
 fstring_format_spec:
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh (approved):
I actually across this yesterday but did not find the time to make a PR. Thanks !

PR #104, opened 2024-10-16
Support Python 3.13

dfremont:
This PR makes the test suite pass under Python 3.13: the only change required is removing a use of the `ASYNC` token in `data/python.gram` (that token no longer exists in Python 3.13). While I was at it I added 3.13 to the CI matrix and updated the Trove classifiers to declare support for Python 3.8-3.13.

I have not fixed the corresponding issue in `data/fullpy.gram` as that grammar has other issues preventing it from working anyway. Nor have I removed `ASYNC` or `AWAIT` from `data/Tokens` since those tokens could still be created when using older versions of Python (and as far as I can tell the file is not used by anything except a broken part of the `show_parse.py` script).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8','3.9','3.10', '3.11', '3.12']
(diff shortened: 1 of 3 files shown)


pablogsal (approved):

pablogsal:
Thanks a lot for the PR @dfremont !

PR #109, opened 2025-03-05
Use more generic python3 shebang

AlexanderS:
The current `python3.8` shebang is all files is very specific. Let's use the more generic `python3` shebang.


MatthieuDartiailh:
This has been addressed as part of #117

PR #111, opened 2025-04-01
Handle multi-line pattern matching expression

### LINKED ISSUE #107: python.gram Rust/C++/Borgo compatible match syntax: parse without delimiters

adsharma:
This is necessary to support an alternative python syntax.

```
# Proposed match_expr using the keyword pmatch
def test(num, num2):
    a = pmatch num:
        1: "One"
        2: "Two"
        3: pmatch num2:
           1: "One"
           2: "Two"
           3: "Three"
        _: "Number not between 1 and 3"
    return a
 ```

Details in the issue. Solution implemented handles it similar to how python's builtin tokenizer handles multi-line expressions involving parens.

Fixes #107

### DIFF
--- a/src/pegen/tokenizer.py
+++ b/src/pegen/tokenizer.py
@@ -28,6 +28,7 @@ class Tokenizer:
         self._verbose = verbose
         self._lines: Dict[int, str] = {}
         self._path = path
+        self._in_match = 0
         if verbose:
             self.report(False, False)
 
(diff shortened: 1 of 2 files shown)


0dminnimda:
I'd be good to add the failing test in the test cases

adsharma:
@0dminnimda the test case fails before the changes and passes after.

PR #112, opened 2026-01-02
Add support for Python 3.14

AlexanderS:
This adds support for Python 3.14.

Python 3.14 introduced support for [PEP 758 - Allow `except` and `except*` expressions without parentheses](https://peps.python.org/pep-0758/). To be able to keep the grammar backwards compatible, it is required to enable/disable some alternatives in the `except_block` (and related) rules depending on the python version. I could not find a way to do so, so I added a "hacky" way to do it with a prefix in the action. If there is any better way to do it, I am open for suggestions.

Apart from the except stuff, there was only an improvement to the error messages on an invalid pattern target.

### DIFF
--- a/data/python.gram
+++ b/data/python.gram
@@ -1047,13 +1047,29 @@ try_stmt[ast.Try]:
 
 except_block[ast.ExceptHandler]:
     | invalid_except_stmt_indent
+    | 'except' e=expressions ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=None, body=b, LOCATIONS) }
+    | 'except' e=expression 'as' t=NAME ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=t.string, body=b, LOCATIONS) }
     | 'except' e=expression t=['as' z=NAME { z.string }] ':' b=block {
+        PY_VERSION < (3, 14);
         ast.ExceptHandler(type=e, name=t, body=b, LOCATIONS) }
     | 'except' ':' b=block { ast.ExceptHandler(type=None, name=None, body=b, LOCATIONS) }
     | invalid_except_stmt
 except_star_block[ast.ExceptHandler]:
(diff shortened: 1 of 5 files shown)


MatthieuDartiailh:
Thanks for starting this. I will try to review ASAP.

data/python.gram
```diff
@@ -1047,13 +1047,29 @@ try_stmt[ast.Try]:
 
 except_block[ast.ExceptHandler]:
     | invalid_except_stmt_indent
+    | 'except' e=expressions ':' b=block {
+        PY_VERSION >= (3, 14);
```
MatthieuDartiailh, line 1051: In the past this was addressed using the check_version method (see decoarators for which we have: `self.check_version((3, 9), "Generic decorator are",  a)` at https://github.com/we-like-parsers/pegen/blob/19158de1bab8bff3181a89a0805ede09e9d87288/data/python.gram#L759-L761.

I think you can use the same solution and avoid extension to the parser logic.

tests/python_parser/test_syntax_error_handling.py
MatthieuDartiailh, line 1: Could you add tests for the newly supported syntaxes ?

data/python.gram
AlexanderS: I tried using `self.check_version`, but it will emit a `SyntaxError` if a rule is matched but the version is to low. I could not image how to use it to selectively enable rules depending on the version. The content of the action (and `check_version` only is an action) can only modify the return value of the rule, as far as I understand. I tried returning `None` or `object()` like in other locations, but the rule will match regardless and will prevent the correct rule from matching.

MatthieuDartiailh: Enabling new rules on newer version works but forbidding rules on old version indeed does not work. However if the generated parser is valid you should be able to keep both rules on recent versions no ? And raising a syntax error is the right behavior since if no old rules match this is what will happen however we provide more detailed error messages.

MatthieuDartiailh:
Superceded by #117

PR #114, opened 2026-01-24
Test on supported python releases

bkmgit:


bkmgit:
covered in https://github.com/we-like-parsers/pegen/pull/112

PR #115, opened 2026-07-17
Fix trailer formatting in python generator

tiago-teixeira5:
The formatting of the trailer string, as documented, was being applied only to the default `MODULE_SUFFIX`, resulting in a custom trailer not having the `{class_name}` being replaced.

From the documentation:

> Note that the trailer is formatted using `.format(class_name=cls_name)` allowing you to reference the created parser in the trailer.


MatthieuDartiailh:
Could you add a test ?

tiago-teixeira5:
> Could you add a test ?

Where would be the best place to add a test?

A story to make sure the generated code includes the class name?

MatthieuDartiailh:
The stories are not run as part of the testsuite, so a test at the root of the tests folder creating a custom parser relying on the feature would be better.

MatthieuDartiailh:
ping @tiago-teixeira5

tiago-teixeira5:
Hi, I had completely forgotten.

Let me ping my alter ego @Wiguwbe

PR #116, opened 2026-09-02
re-add hash function to Cut class

jtaylor05:
Because the Cut class has a __eq__ dunder definition, it automatically nulls the __hash__ object dunder. This makes it impossible to hash for dictionary or sets.

This patch adds back in the base object.__hash__ definition for the Cut class, while leaving the __eq__ dunder unmodified.

Example use-case:
A subclass of the GrammarVisitor which only visits each node once even if there are recursive nodes.


MatthieuDartiailh:
Could you rebase on main ?

jtaylor05:
Should be rebased now.

MatthieuDartiailh:
What bothers me a bit here is that all instances of Cut compare equal so we end up with different hashes for equal objects which does not seem meaningful.

jtaylor05:
I think that is a valid concern. The hash function could be changed so that all Cut objects share the same hash. Perhaps we hash one statically stored cut instance that all cuts share. 
The more permanent solution would be to condense equal value instances of the grammar tree so that they literally are the same instance. This could be done for StringLeaf and NameLeaf classes as well because they are also Leaf nodes.

MatthieuDartiailh:
Yes turning those classes into actual singleton may make more sense. Sadly I do not the history of why it was not done initially.

PR #117, opened 2026-09-09
Add support for Python 3.15

maxyz:
This pr is based on the work done by Alexander Sulfrian in pr#112.

I added some fixes that were required to satisfy the more strict validations in python 3.15.
And added the support for the new python 3.15 changes.

The Python 3.15 support introduces:
 - lazy imports, detailed in [PEP 810](https://peps.python.org/pep-0810/): Explicit lazy imports
 - updates of message phrasing in exceptions [1]
 - comprehension unpacking, detailed in [PEP 798](https://peps.python.org/pep-0798/): Unpacking in comprehensions

[1]: https://docs.python.org/3.15/whatsnew/3.15.html#improved-error-messages

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8','3.9','3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.8','3.9','3.10', '3.11', '3.12', '3.13', '3.14']
     steps:
     - uses: actions/checkout@v4
     - name: Get history and tags for SCM versioning to work
--- a/data/python.gram
+++ b/data/python.gram
@@ -1047,13 +1047,29 @@ try_stmt[ast.Try]:
 
 except_block[ast.ExceptHandler]:
     | invalid_except_stmt_indent
+    | 'except' e=expressions ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=None, body=b, LOCATIONS) }
+    | 'except' e=expression 'as' t=NAME ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=t.string, body=b, LOCATIONS) }
     | 'except' e=expression t=['as' z=NAME { z.string }] ':' b=block {
+        PY_VERSION < (3, 14);
         ast.ExceptHandler(type=e, name=t, body=b, LOCATIONS) }
     | 'except' ':' b=block { ast.ExceptHandler(type=None, name=None, body=b, LOCATIONS) }
     | invalid_except_stmt
 except_star_block[ast.ExceptHandler]:
     | invalid_except_star_stmt_indent
+    | 'except' '*' e=expressions ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=None, body=b, LOCATIONS)
+     }
+    | 'except' '*' e=expression 'as' t=NAME ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=t.string, body=b, LOCATIONS)
+     }
     | 'except' '*' e=expression t=['as' z=NAME { z.string }] ':' b=block {
+        PY_VERSION < (3, 14);
         ast.ExceptHandler(type=e, name=t, body=b, LOCATIONS)
      }
     | invalid_except_stmt
@@ -2256,7 +2272,12 @@ invalid_try_stmt[NoReturn]:
         )
      }
 invalid_except_stmt[None]:
+    | 'except' '*'? a=expression ',' expressions 'as' NAME ':' {
+        PY_VERSION >= (3, 14);
+        self.raise_syntax_error_starting_from("multiple exception types must be parenthesized when using 'as'", a)
+     }
     | 'except' '*'? a=expression ',' expressions ['as' NAME ] ':' {
+        PY_VERSION < (3, 14);
         self.raise_syntax_error_starting_from("multiple exception types must be parenthesized", a)
      }
     | a='except' '*'? expression ['as' NAME ] NEWLINE { self.raise_syntax_error("expected ':'") }
--- a/data/python.gram
+++ b/data/python.gram
@@ -2315,8 +2315,11 @@ invalid_as_pattern[NoReturn]:
     | or_pattern 'as' a="_" {
         self.raise_syntax_error_known_location("cannot use '_' as a target", a)
      }
-    | or_pattern 'as' !NAME a=expression {
-        self.raise_syntax_error_known_location("invalid pattern target", a)
+    | or_pattern 'as' a=expression {
+        self.raise_syntax_error_known_location(
+            f"cannot use {self.get_expr_name(a)} as pattern target", a)
+        if self.py_version >= (3, 14)
+        else self.raise_syntax_error_known_location("invalid pattern target", a)
      }
 invalid_class_pattern[NoReturn]:
     | name_or_attr '(' a=invalid_class_argument_pattern  {
--- a/src/pegen/python_generator.py
+++ b/src/pegen/python_generator.py
@@ -397,6 +397,7 @@ class PythonParserGenerator(ParserGenerator, GrammarVisitor):
         locations = False
         unreachable = False
         used = None
+        version_check = None
         if action:
             # Replace magic name in the action rule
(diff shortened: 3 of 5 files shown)


maxyz:
@MatthieuDartiailh could you please take a look?

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8','3.9','3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.8','3.9','3.10', '3.11', '3.12', '3.13', '3.14', '3.15']
```
MatthieuDartiailh:
```suggestion
        python-version: ['3.8','3.9','3.10', '3.11', '3.12', '3.13', '3.14', '3.15-dev']
```

MatthieuDartiailh: 3.15 is not yet released

data/python.gram
```diff
@@ -1047,13 +1055,29 @@ try_stmt[ast.Try]:
 
 except_block[ast.ExceptHandler]:
     | invalid_except_stmt_indent
+    | 'except' e=expressions ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=None, body=b, LOCATIONS) }
+    | 'except' e=expression 'as' t=NAME ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=t.string, body=b, LOCATIONS) }
```
MatthieuDartiailh: Those 2 can be concatenated in 1 using the pattern used in the pre 3.14 rule.

data/python.gram
```diff
@@ -1047,13 +1055,29 @@ try_stmt[ast.Try]:
...
 except_star_block[ast.ExceptHandler]:
     | invalid_except_star_stmt_indent
+    | 'except' '*' e=expressions ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=None, body=b, LOCATIONS)
+     }
+    | 'except' '*' e=expression 'as' t=NAME ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=t.string, body=b, LOCATIONS)
+     }
```
MatthieuDartiailh: Those 2 can be concatenated in 1 using the pattern used in the pre 3.14 rule.

src/pegen/python_generator.py
MatthieuDartiailh, line 1: I am not exactly convinced by that. It makes the grammar a bit less verbose but a bit harder to reason about IMO.

MatthieuDartiailh (commented):
Thanks for the PR. Before merging I would like to see some extra tests:
- tests related to 3.14 syntax (as requested in #112 review)
- test for wildcard pattern (and wildcard in star pattern) in match since we were building the wrong node
- support for explicit lazy import and related tests

MatthieuDartiailh:
Looks nice. While we are at it, can you drop 3.8 ? 

Also I am still not fully onboard with PY_VERSION magic. Could you make a case for it ?

maxyz:
> Looks nice. While we are at it, can you drop 3.8 ?

I pushed a change that drops 3.8. I also updated the tox configuration but I don't use it, so you might need to tweak it.

> Also I am still not fully onboard with PY_VERSION magic. Could you make a case for it ?

This came from the previous patch I used it for convenience, I have no strong opinions about it, if you have a better alternative let me know.

tests/test_pegen.py
MatthieuDartiailh, line 1: Why did you need to remove the locations ?

tox.ini
```diff
@@ -1,11 +1,10 @@
 [tox]
-envlist = py38, py39, py310, py311, py312, py38-cov, py39-cov, py310-cov, py311-cov, py312-cov, lint, regen, docs
+envlist = py39, py310, py311, py312, py39-cov, py310-cov, py311-cov, py312-cov, lint, regen, docs
```
MatthieuDartiailh, line 2:
```suggestion
envlist = py39, py310, py311, py312, py313, py314, py315, py39-cov, py310-cov, py311-cov, py312-cov, py313-cov, py314-cov, py315-cov, lint, regen, docs
```

tox.ini
```diff
@@ -1,11 +1,10 @@
...
 isolated_build = true
 
 [gh-actions]
 python =
-    3.8: py38-cov, lint, regen, docs
-    3.9: py39-cov
+    3.9: py39-cov, lint, regen, docs
     3.10: py310-cov
     3.11: py311-cov
     3.12: py312-cov
```
MatthieuDartiailh, line 10:
```suggestion
     3.12: py312-cov
     3.13: py313-cov
     3.14: py314-cov
     3.15: py315-cov
```

data/python.gram
```diff
@@ -713,14 +713,31 @@ import_stmt[ast.Import]:
 # Import statements
 # -----------------
 
-import_name[ast.Import]: 'import' a=dotted_as_names { ast.Import(names=a, LOCATIONS) }
+import_name[ast.Import]: lazy="lazy"? 'import' a=dotted_as_names {
+    ast.Import(names=a, is_lazy=1 if lazy else 0, LOCATIONS)
+    if sys.version_info >= (3, 15) else
+    self.check_version((3, 15), "Lazy imports are", ast.Import(names=a, LOCATIONS))
+    if lazy else
+    ast.Import(names=a, LOCATIONS)
```
MatthieuDartiailh, line 721: I really tried to find a nicer way to express this but failed...

data/python.gram
```diff
@@ -1047,12 +1064,19 @@ try_stmt[ast.Try]:
 
 except_block[ast.ExceptHandler]:
     | invalid_except_stmt_indent
+    | 'except' e=expressions ':' b=block {
+        PY_VERSION >= (3, 14);
+        ast.ExceptHandler(type=e, name=None, body=b, LOCATIONS) }
     | 'except' e=expression t=['as' z=NAME { z.string }] ':' b=block {
         ast.ExceptHandler(type=e, name=t, body=b, LOCATIONS) }
     | 'except' ':' b=block { ast.ExceptHandler(type=None, name=None, body=b, LOCATIONS) }
```
MatthieuDartiailh, line 1072:
```suggestion
     | 'except' e=expression t=['as' z=NAME { z.string }] ':' b=block {
         ast.ExceptHandler(type=e, name=t, body=b, LOCATIONS) }
     | 'except' ':' b=block { ast.ExceptHandler(type=None, name=None, body=b, LOCATIONS) }
    | 'except' e=expressions ':' b=block {
        ast.ExceptHandler(type=e, name=None, body=b, LOCATIONS) if self.py_version >= (3, 14) else None }
```
I did not get a chance to test but I think this should work. Could you test and apply the pattern in other locations ?

data/python.gram
```diff
@@ -1762,21 +1786,38 @@ for_if_clause[ast.comprehension]:
     | invalid_for_target
 
 listcomp[ast.ListComp]:
-    | '[' a=named_expression b=for_if_clauses ']' { ast.ListComp(elt=a, generators=b, LOCATIONS) }
+    | '[' a=(starred_expression | named_expression) b=for_if_clauses ']' {
+        PY_VERSION >= (3, 15);
+        ast.ListComp(elt=a, generators=b, LOCATIONS) }
+    | '[' a=named_expression b=for_if_clauses ']' {
+        PY_VERSION < (3, 15);
+        ast.ListComp(elt=a, generators=b, LOCATIONS) }
```
MatthieuDartiailh, line 1794: I think you can invert the rules and remove the version check on the first one. It is more specific than the 3.15 rules so except as an optimization it is not necessary.

MatthieuDartiailh (commented):
Hopefully a last batch of comments.

Could you also add a changelog entry ?

tests/test_pegen.py
maxyz: Because ast.Expression no longer accepts unexpected kwargs, this was a warning in 3.14 and an error in 3.15.

MatthieuDartiailh: Thanks for the clarification

MatthieuDartiailh (approved):

maxyz:
I reformatted the code to match black, now all the tests are green. Please merge if you think that's ok.

MatthieuDartiailh:
Everything looks nice. Could just add an entry in the release notes ?

maxyz:
> Everything looks nice. Could just add an entry in the release notes ?

Updated, and merged with main. PTAL

MatthieuDartiailh (approved):

MatthieuDartiailh:
Thanks !

PR #118, opened 2026-09-10
Fix f-string conversion specifiers on Python 3.12+

Souravrajvi0:
## Summary

Fixes f-string conversion specifiers (`!r`, `!s`, `!a`) on Python 3.12+, where `tokenize.TokenInfo` uses `start`/`end` positions instead of `lineno`/`col_offset`.

- Use `mark.end != name.start` instead of comparing `lineno`/`col_offset` in `check_fstring_conversion`
- Return `ord()` of the conversion character instead of the `TokenInfo` object
- Use the integer conversion value directly in `ast.FormattedValue` instead of `conversion.decode()[0]`

## Context

This is the same fix merged upstream in [BerkeleyLearnVerify/Scenic#501](https://github.com/BerkeleyLearnVerify/Scenic/pull/501), suggested by @dfremont since Scenic's grammar is derived from Pegen's Python grammar.

## Test plan

- [x] Verified `!r`, `!s`, `!a` produce AST matching CPython
- [x] Verified invalid conversion (`!z`) raises the expected `SyntaxError`
- [x] Added regression cases to `tests/python_parser/data/fstrings.py` and `test_syntax_error_handling.py`


tests/python_parser/test_syntax_error_handling.py
```diff
@@ -162,6 +162,12 @@ def parse_invalid_syntax(
             (3, 12) if sys.version_info >= (3, 12) else (3, 7),
             (3, 13) if sys.version_info >= (3, 12) else (3, 8),
         ),
+        (
+            'f"{x!z}"',
+            "f-string: invalid conversion character 'z': expected 's', 'r', or 'a'",
+            (1, 10),
+            (1, 11),
```
MatthieuDartiailh:
```suggestion
            (1, 6),
            (1, 7),
```

MatthieuDartiailh (commented):
Looks great ! A small suggestion to fix the CI (I let you double the values).

Souravrajvi0:
> Looks great ! A small suggestion to fix the CI (I let you double the values).

Fixed  updated to (1, 6) / (1, 7). Thanks

MatthieuDartiailh:
Could you add a changelog entry?

With it I think we would be all set.

Souravrajvi0:
> Could you add a changelog entry?
> 
> With it I think we would be all set.

Hi change log is updated, ready for the final review

MatthieuDartiailh (approved):

MatthieuDartiailh:
Thanks. The release may end up being 0.4.0 to include 3.14 and 3.15 support.

MatthieuDartiailh:
You have a mismatch between code and test for an error message.

